PR #303, opened 2017-05-11
Update roadmap.md

shaggygi:
Fix misspelling.


crutkas:
thanks @shaggygi

crutkas:
@shaggygi thanks for the fix, next time it would be great to open an issue as well so we can track the fix.  Thanks!

PR #309, opened 2017-05-11
Make sure OK is always OK, and not Ok

mattleibow:
The inconsistency gets to me deep down ;)


crutkas:
This is against master, not Dev, can't approve this.

crutkas:
And the nice twitter people showed me how to change the base branch :)

crutkas:
Thanks @mattleibow for the fix!

PR #356, opened 2017-05-17
Fix versioning issues 

ralarcon:
Microsoft.TemplateEngine versioning issues and other issues


ralarcon:
All builds are passing... The check does not reflect it.

Issue #350, opened 2017-05-16
App mode Light/Dark/Windows default


myokeeh:
It would be nice to have best practice Light/Dark/Windows default app mode settings baked into WTS to promote consistency.

This one's from the Maps app
![capture](https://cloud.githubusercontent.com/assets/3663297/26132870/032d9b40-3a58-11e7-9f0b-a7c2051652dc.PNG)

...among many!

crutkas:
Fair, settings page will give you theme support but i do see that we don't respect the default.

would this be something you'd be interested in doing a PR for?  :)

crutkas:
For the notes, i think this should be done by default for all apps

degant:
Can I give this a shot? :)

crutkas:
@degant 100%.  Remember to branch of dev.  I think this should be a pretty straight forward feature.  The Theme setting on the setting page may want to default to the OS theme then as well.

Studio384:
I would say that there should also be taken a look at making this instant and not require users to restart the app. Skype does this, and if the theme is set to use the system color and it changes it works just fine, so why not allow users to switch the theme without restarting?

mrlacey:
@Studio384 we will not be requiring a restart to change theme. (we don't currently either.)
Making it work without a restart is part of the challenge that's made this slow to implement.

crutkas:
@studio384 I shared with Matt a POC. I think I have a working version.

Studio384:
@mrlacey Had forgotten that WTS already does this without restart. The screenshot confused me. :)

crutkas:
ok, think i got a near final POC.  Code behind but does everything, needs maybe a bit more TLC on setting page look feel but is functional now.

https://github.com/crutkas/Playground/tree/master/Issue350

qbss:
I'm looking forward to this becoming active and maybe someone could correct this related minor issue (# 903) during the process.

mrlacey:
The old XAML has been left commented out in the SettingsPage.xaml file and needs to be removed.

crutkas:
Crap.  Will do today.  Good catch.

mrlacey:
Thre's also a bug for Pivot&Tabs with CodeBehind - going to the settings tab resets to default

Repro:

- With Windows default theme set to Light on the PC/device
- New project with Pivot&Tabs and CodeBehind
- Add a settings page
- Run the generated app
- Go to the settings page and switch to dark theme
- Close app
[...]

PR #360, opened 2017-05-17
Feature #350 Including support for default Windows Theme

degant:
PR for feature #350 to support **Windows Default theme** in addition to Light and Dark themes. This mode will set the app's theme as the default windows theme which can be controlled directly from 'Color Settings'. 

- Switching to `RadioButton` instead of `ToggleSwitch` since we have 3 modes: Light, Dark and Windows default
- Removed `RequestedTheme='Light'` from Application tag in `App.xaml` since XAML doesn't support a default mode so to respect the system's theme, this has to be removed.
- In `ShellNavigationItem`, brushes `SystemControlForegroundAltHighBrush` and `SystemControlForegroundBaseHighBrush` now behave based on the default windows setting since `RequestedTheme` isn't set to Light in the App.xaml. As a result, they can't be used when the theme is set to Light or Dark _in the app_. So manually setting them to `White` and `Black` (File: `ShellNavigationItem_postaction.cs`). Couldn't find a better approach for this

Screenshot:

### DIFF
--- a/templates/Pages/Settings.CodeBehind/Services/ThemeSelectorService.cs
+++ b/templates/Pages/Settings.CodeBehind/Services/ThemeSelectorService.cs
@@ -13,6 +13,8 @@ namespace Param_RootNamespace.Services
         public static event EventHandler<ElementTheme> OnThemeChanged = delegate { };
 
         public static bool IsLightThemeEnabled => Theme == ElementTheme.Light;
+        public static bool IsDarkThemeEnabled => Theme == ElementTheme.Dark;
+        public static bool IsDefaultThemeEnabled => Theme == ElementTheme.Default;
         public static ElementTheme Theme { get; set; }
 
         public static async Task InitializeAsync()
@@ -20,16 +22,11 @@ namespace Param_RootNamespace.Services
             Theme = await LoadThemeFromSettingsAsync();
         }
 
-        public static async Task SwitchThemeAsync()
+        public static async Task SetThemeAsync(string themeName)
         {
-            if (Theme == ElementTheme.Dark)
-            {
-                await SetThemeAsync(ElementTheme.Light);
-            }
-            else
-            {
-                await SetThemeAsync(ElementTheme.Dark);
-            }
+            ElementTheme theme;
+            Enum.TryParse<ElementTheme>(themeName, out theme);
+            await SetThemeAsync(theme);
         }
 
         public static async Task SetThemeAsync(ElementTheme theme)
@@ -51,13 +48,9 @@ namespace Param_RootNamespace.Services
 
         private static async Task<ElementTheme> LoadThemeFromSettingsAsync()
         {
-            ElementTheme cacheTheme = ElementTheme.Light;
+            ElementTheme cacheTheme = ElementTheme.Default;
             string themeName = await ApplicationData.Current.LocalSettings.ReadAsync<string>(SettingsKey);
-            if (String.IsNullOrEmpty(themeName))
-            {
-                cacheTheme = Application.Current.RequestedTheme == ApplicationTheme.Dark ? ElementTheme.Dark : ElementTheme.Light;
-            }
-            else
+            if (!String.IsNullOrEmpty(themeName))
             {
                 Enum.TryParse<ElementTheme>(themeName, out cacheTheme);
             }
--- a/templates/Pages/Settings.CodeBehind/Strings/en-us/Resources_postaction.resw
+++ b/templates/Pages/Settings.CodeBehind/Strings/en-us/Resources_postaction.resw
@@ -7,11 +7,14 @@
   <data name="SettingsPage_Theme.Text" xml:space="preserve">
     <value>Theme</value>
   </data>
-  <data name="SettingsPage_ThemeToggle.OnContent" xml:space="preserve">  
+  <data name="SettingsPage_Theme_Light.Content" xml:space="preserve">  
     <value>Light</value>
   </data>
-  <data name="SettingsPage_ThemeToggle.OffContent" xml:space="preserve">    
+  <data name="SettingsPage_Theme_Dark.Content" xml:space="preserve">    
     <value>Dark</value>
+  </data>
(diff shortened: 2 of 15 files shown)


crutkas:
Solid.  

- I think a drop down may be better here to conserve space.  
- The icons don't flip if a user changes the background color but everything else does.

I'm going to roll this into its own branch.

crutkas:
How the shell navigation post action color was being handled is why the icons don't flip i believe.  You shouldn't be hardcoding colors.  Use theme based brushes.

crutkas:
Another reason why we should always defer to system brushes is incase someone is on high contrast

degant:
Thanks for the feedback @crutkas 
> - I think a drop down may be better here to conserve space.

Should I change it to a drop-down instead? I was trying to follow the Maps app so picked radio buttons

> - The icons don't flip if a user changes the background color but everything else does.

With the app's theme set to **Windows default** and the app open, you are trying to change the system's default app mode (Light or Dark). And if a Navigation app is selected then the navigation item's foreground color doesn't change. Is that the scenario you are referring?

I observed the same behavior and was not satisfied with using hardcoded Colors instead of theme colors but couldn't come up with a better approach that tackled all scenarios. Currently I see too major problems with the implementation:

1. When the system theme is flipped, then the navigation pane doesn't reflect the same change (in realtime) so the icons and text appear don't appear at all
1. Hardcoded `Color` values are used for shell navigation, which should be avoided

**Regarding the 1st problem: Navigation Item colors don't flip when system theme is updated:**
- `SystemControlForegroundBaseHighBrush` represents the System's foreground color so if the color is set in XAML using `ThemeResource` like this: `{ThemeResource SystemControlForegroundBaseHighBrush}` then it works based on the system's app mode (Light or Dark) if the Windows default option is selected in app theme, and also based on the app's theme (if app theme is set to Light or Dark)
- Unfortunately if the same thing is done via code instead of XAML, i.e by binding to `Application.Current.Resources["SystemControlForegroundBaseHighBrush"] as SolidColorBrush` then every time the Windows default mode is changed from Light to Dark, the app doesn't reflect the same.
- Using XAML `ThemeResource` would fix this problem but we can't do that since we need to determine color based on whether the navigation item is selected or not
- Another observation is that when system theme is updated (or high contrast setting is applied), then all XAML resources of the app get updated in accordance with the changes. But since this one is bound to a `SolidColorBrush` it doesn't reflect changes. However, if the user toggled the app theme then all changes reflect appropriately with the navigation item colors. 
- A solution that I could think of was to to trigger the `ThemeSelectorService.OnThemeChanged` everytime system color settings are modified so that the navigation item's color's are correctly set. That would mean subscribing to `UISettings.ColorValuesChanged` event handler and then updating our colors everytime (only for the navigation pane).

**Regarding the 2nd problem: Remove hardcoding of colors:**
- The implementation for Navigation Items requires binding the `Foreground` property so that the selected one's color is bound to `SystemControlForegroundAccentBrush` and the rest to `SystemControlForegroundBaseHighBrush`. On changing the app theme in the app itself, the navigation items determine the foreground colors once again and then accessing `SystemControlForegroundBaseHighBrush` from `Resources` always returns the foreground color that has been set in the system's color settings. 
_**Example:** If windows is set to **dark** app mode, and the app is set to **light**, then accessing the `SystemControlForegroundBaseHighBrush` color from code behind returns white whereas our app would need a dark color for foreground._ 
- So instead I had to set it to `Black` when the Light theme was enabled from the app since most of the theme based brushes were returning colors based on the system color settings instead of the app. The only slightly dirty solution that I could come up to avoid using White & Black colors directly was to use `SystemControlForegroundAltHighBrush` _when the system's theme was different from the app's theme_. Now the only issue here is how to determine the system's theme. This is the only way I could do it, which isn't very clean because of the comparison with `Colors.Black`.

      var result = Application.Current.Resources["SystemControlForegroundBaseHighBrush"] as SolidColorBrush;
      UISettings uiSettings = new UISettings();
      var backgroundColor = uiSettings.GetColorValue(UIColorType.Background);
      bool isDark = backgroundColor == Colors.Black;

      // If app theme is different from system theme, then set foreground is AltHighBrush instead
      if ((Services.ThemeSelectorService.IsLightThemeEnabled && isDark) ||
          (Services.ThemeSelectorService.IsDarkThemeEnabled && !isDark))
      {
          result = Application.Current.Resources["SystemControlForegroundAltHighBrush"] as SolidColorBrush;
      }
- Does this feel like an acceptable solution to you? Is there a different way to check system app mode?

I feel I might be missing something straightforward while trying to access brushes for each theme the right way. Appreciate if someone can point me in the right direction, or suggest a better, cleaner approach :)

Apologies for the really long and confusing description. Tried my best to convey all observations in detail so that a better solution or simpler approach can be found.

degant:
Hey @crutkas, any thoughts on the above?
Should I be moving this discussion to the issue instead, since this PR is marked as closed?

crutkas:
Sure. Or create a new PR. I did create a new branch for this.

degant:
Yes, you mentioned the new branch. I was mostly looking for help with the approach. I've already outlined the issues I've ran into and the slightly hacky solutions. So not sure if I have anything for a PR. Though I could do the drop-down thing if that's okay? But we would still be left with 2 problems in the design on which I'm seeking guidance.

crutkas:
The branch is: https://github.com/Microsoft/WindowsTemplateStudio/tree/Issue350_DefaultThemeSupport

@mrlacey has a great two bits of feedback here: 
 https://github.com/Microsoft/WindowsTemplateStudio/compare/Issue350_DefaultThemeSupport

- I would do a dropdown but has to be localized.
- I would need to see the code in a runnable state to see.

degant:
Hi @crutkas, @mrlacey thanks, will work on a PR with dropdown tomorrow. The code right now (with radio buttons) is in a runnable state and passing all template generation cases. I've manually tested quite a few permutations and combinations too.

But I'm hoping to get some help on the two issues I've outlined above in the slightly long post. Appreciate if you could take a look at them. Thanks!

Issue #343, opened 2017-05-16
Templates not shown in Blend


ridomin:
Can you enable the WTS wizard from Blend File->New Project ?

mrlacey:
It shows up when searching

![image](https://cloud.githubusercontent.com/assets/189547/26123758/e7266034-3a30-11e7-8014-5dcde77988a8.png)

Install reports success but then it doesn't show up in the New Project list

crutkas:
Happily will accept a PR for someone to add it in.

mrlacey:
According to a comment, by someone on the VS IDE team, on a related issue [here](https://github.com/github/VisualStudio/issues/951) extensions in Blend require whitelisting.
If we really want this, @crutkas can you find out more about whitelisting?

mgoertz-msft:
@crutkas I have a local fix for this. How do I create a PR? I cannot push my local branch and I cannot fork the repo either. Do you need to add me to this project first?

mgoertz-msft:
PR #361 has the fix for this issue.

crutkas:
in dev branch now

PR #361, opened 2017-05-17
Templates not shown in Blend #343

mgoertz-msft:
Install extension for all users so it can be seen by Blend
Add Blend AppIdFilter to template so that it shows up in Blend.


crutkas:
Fix for #343

Issue #311, opened 2017-05-12
Better support / Need more specific error for offline use


mrlacey:
"Error updating templates. See output for more details." is not helpful enough. Can we add a more specific error if unable to download templates as offline.
Would also be good to add a retry option without having to close the wizard.

Repro:
- Install VSIX
- Go offline
- File > New > WTS

nothing downloads and have to close our dialog and the Choose Project type dialog before can see the output window. :(

crutkas:
Agreed, we need better offline support.  If you have a cached version, it should continue to run.

mrlacey:
Changing the milestone on this as a real report of this problem was marked for addressing as a critical bug fix.

ralarcon:
I'm assiging this to myself as I will be working on it.

Issue #277, opened 2017-05-09
Templates are not downloaded in background


ralarcon:
After the UI redesign the ability to download the templates in background has been lost, that is, the user cannot continue working if a compatible template content is locally available while new content is being donwloaded.

It looks it is something related to the UI and when it consider you can continue working.

mrlacey:
What's the rationale behind this bug? If someone is online and there are newer templates available don't we want them to use the newer ones. They should only fall back to cached older ones if can't check for new templates or there are problems during the download.

ralarcon:
You are right, the thing is that everything was thought to make the downloads in background so if you have available content you can continue working without waiting to get the latest version of templates. 

It is not really important if you have good internet connection but we are showing some exceptions being tracked in the telemetry regarding timeouts in the download or not connectivity issues.

This now has been fixed and, by the way, we have added support to work in off-line mode even the first time you launch the wizard.

mrlacey:
So, if 
1. Person installs the VSIX
2. Person uses it to create an app
3. New version of template is released with bug fix
4. Person goes to create another app they still have to use old, buggy templates
5. Person then goes to create a third app then they get the templates with the fix

This concerns me. 
For the sake of a check for new content when the tool loads we stop people having access to the latest and greatest templates.

Typical developers don't create new apps very often. They could end up using templates that are weeks out of date.

Yes, fallback to cached version if offline or there's a problem getting the new template.
Allow a person to skip the check if they really want too.
Don't check more often than every N hours.
But, do try and get people using the latest version of templates by default.


If connectivity is an issue in terms of the time to retrieve and download the templates, then those people are going to be in for a real shock when nuget package restore kicks in when they try and build the app.

ralarcon:
We can allow users to search explicitly for new templates and / or we can allow users to force the refresh of recent downloaded templates... if so, we can open a different issue to track this.

Issue #332, opened 2017-05-13
Localize the wizard


crutkas:
Visual Studio is used by the world and our wizard should reflect that.  If someone has VS set to a different language, our wizard should respect that.

this is a different issue that #46.  #46 is about making the code product localized content that is generated.

crutkas:
FYI, we do have code debt for warnings.

ralarcon:
Latest resources ready to be localized. Waiting to get them back integrated.

ralarcon:
Moving to 1.5 as we do not have the localized files jet.

sibille:
Verified on dev-nightly:

Assistent: 0.14.17307.65003
Vorlagen: 0.14.17307.4

PR #369, opened 2017-05-18
Fixes for #311 and #277

ralarcon:


code/src/Core/Locations/RemoteTemplatesSource.cs
```diff
@@ -63,11 +72,9 @@ private static void DownloadContent(string sourceUrl, string file)
             }
             catch (Exception ex)
             {
-                string msg = "The templates content can't be downloaded.";
-
-                AppHealth.Current.Exception.TrackAsync(ex, msg).FireAndForget();
-
-                throw;
+                string msg = $"Templates content can't be downloaded right now, we will try it later.";
```
mrlacey: Message should be localized

mrlacey: This text implies that the system will automatically try and download it later. Is this really the case?

code/src/UI/ViewModels/MainViewModel.cs
```diff
@@ -153,14 +155,14 @@ public MainViewModel(MainView mainView)
...
-                Status = new StatusViewModel(StatusType.Information, StringRes.ErrorSync);
+                Status = new StatusViewModel(StatusType.Information, StringRes.ErrorSync, true);
 
                 await AppHealth.Current.Error.TrackAsync(ex.ToString());
                 await AppHealth.Current.Exception.TrackAsync(ex);
             }
             finally
             {
-                Current.LoadingContentVisibility = Visibility.Collapsed;                
+                //Current.LoadingContentVisibility = Visibility.Collapsed;
```
mrlacey: why not just remove the finally block entirely?

ralarcon: missing... fixing right now.

code/src/UI/Views/MainView.xaml
```diff
@@ -69,10 +69,9 @@
             <Grid Visibility="{Binding LoadingContentVisibility}">
                 <TextBlock Style="{StaticResource TextMainLoadingStyle}" Text="{x:Static res:StringRes.LoadingTemplatesText}" />
             </Grid>
-            <Grid Visibility="{Binding NoContentVisibility}">
-                <TextBlock Style="{StaticResource TextMainLoadingStyle}" Text="{x:Static res:StringRes.NoContentText}" />
+            <Grid Visibility="{Binding LoadedContentVisibility}">
```
mrlacey: Could you not just do away with this outer grid and use the visibility binding on the inner textblock to control its display? 
Nested single elements with different visibility bindings are hard to debug and maintain.

code/src/Core/Locations/RemoteTemplatesSource.cs
ralarcon: yes, it is the case, it will try to download it again next time WTS is used (if have passed at least 120 mins from last download)

ralarcon: Yes, I didn't have done because it is in the core and it will be shown only in the log (output)...

mrlacey: #332 will require this to be localized. Better to do it now than have to come back and do it later

mrlacey: That's not how I understand the message. Your comment above contains lots of information that isn't communicated. "we will try it again later" doesn't say when. Nor that there are conditions about trying again.

ralarcon: It is just a log message to indicate that the download didn't worked and will be try later... Suggestions are welcomed.

crutkas:
for retry, create new issue.  I don't want to hold off.

ralarcon:
Retry option looks like another feature for me, even we can set a place to request the donwload explictly. Regarding the local testing you will need to change from debug to release to use the RemoteTemplatesSource within the VSIX locally. This is a critical update for off-line support and avoid being block while downloading everytime. A review on this is required to better handle the local testing.

code/src/Core/Locations/RemoteTemplatesSource.cs
crutkas: This is a blocking bug right now.  I hear you but disagree as that is adding in additional work here.

Issue #84, opened 2017-03-28
Folders created on disk even if Wizard is cancelled or backed out


mrlacey:
**Repro**
- From the New Project window select the template based project.
- Default app name will be like "App5"
- Press OK
- Get taken to the Wizard.
- Press Cancel
- Back on the New Project window

**Actual**
- Default app name is now "App6" (one number higher)
- Empty folders have been created on disk based on the project (and solution) name

**Expected**
- default name still "App5" and nothing created on disk

Nothing should be created on disk until the wizard is complete.

ralarcon:
Verified in nightly build 0.14.17146.65001

PR #395, opened 2017-05-25
Delete empty folder created by Visual Studio when cancelling wizard 

sibille:
Delete empty folder created by Visual Studio when cancelling wizard  without project creation #84


mrlacey:
I'm a bit concerned that this deletes the folder on any exception.
I understand if it's on a WizardBackoutException but for *any* exception (even a WizardCancelledException) then we may lose useful debugging information around what caused the error.

Also, does the delete need to be recursive? Why would items have been created before someone cancels the wizard? Or is this just to tidy up after an error?

sibille:
I'll change it to delete the folder only on WizardBackoutException. Although exceptions are tracked via telemetry, the failed generation could be valueable. 
Delete is recursive to delete solution and projectfolder (in case you checked create directory for solution)

Issue #399, opened 2017-05-25
Improve Naming Infer to apply validation steps independently


sibille:
Naming.Infer does the following: 
- Replace invalid chars
- Check against reserved names
- Check against existing names

Separate this into various steps that can be applied independently. Check also DefaultInstancenames from singleInstance templates are not used in templates that allow name edition. (f.ex a Page cannot be named LiveTile) to avoid collision.

ralarcon:
Verified in nightly build 0.14.17146.65001

PR #400, opened 2017-05-25
Refactor naming to make namig agnostic of templates

sibille:
#399


code/src/Core/Naming/ReservedNameValidator.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
```
crutkas: This needs a header

code/src/Core/Naming/ExistingNamesValidator.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Microsoft.Templates.Core;
```
crutkas: This needs a header

code/src/Core/Naming/Validator.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
crutkas: This needs a header

sibille:
I'll add them in, thanks!!

Issue #417, opened 2017-05-29
AssemblyInfo shows year of 1976 instead of current year.


crutkas:
File->new blank app
[assembly: AssemblyCopyright("Copyright ©  2017")]

File->WTS->blank + MVVM Light
[assembly: AssemblyCopyright("Copyright ©  1976")]

mrlacey:
1976 is a placeholder that should be replaced by the current year.
Config for this is in WTS\templates\Projects\Default\.template.config\template.json

sibille:
Fixed with PR #419

ralarcon:
Verified in latest nightly.

PR #420, opened 2017-05-29
Fixed issue #417: copyrightYear parameter (must be under symbols)

sibille:
Fixed issue #417: copyrightYear parameter (must be under symbols)

### DIFF
--- a/templates/Projects/Blank/.template.config/template.json
+++ b/templates/Projects/Blank/.template.config/template.json
@@ -38,14 +38,14 @@
     {
       "type": "parameter",
       "replaces": "Param_Templates.Version"
-    }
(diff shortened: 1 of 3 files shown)


crutkas:
Why is this getting merged into master?

ralarcon:
This is a hotfix to correct #417 just in case we want to release a new version of templates. 

It is already fixed in dev but the project templates definition has been changed (merged from three projects to just one) and it will not needed to be merged back to dev.

crutkas:
I don't think we need to do a hot fix for this.

ralarcon:
Removing this PR as we will not deploying a hotfix for this.

Issue #403, opened 2017-05-25
Default common dataset for sample data


crutkas:
With the pending chart and data grid, a richer (maybe generated) default data model may be needed.  This would prevent each page from generating its own.

crutkas:
One thought is to use some nasa asteroid data like they did for the custom control talk at Build 2017

mrlacey:
Maybe we have a SampleDataService to provide this. *Possibly* using it as a guide for whatever we do with other data sources...

crutkas:
not a bad idea.  I just don't want to over complicate stuff but still have enough base data to be useful and not have different sample data for different things.

mrlacey:
A single SampleDataService might be a good single place to collect all the different data used to prepopulate pages. It would remove the duplication when adding multiple copies of the same page and would be an isolated thing that could easily (and should be) removed.
Need to think about the best way to add a file to a project only if it is used by one or other items and those items can each merge with it.
Might require a dependency that doesn't show in the wizard and that isn't something we yet support. But should be easy to add.

ralarcon:
Yes, SampleDataService will work being the base sample data for everything needed. Some time ago we had in mind to allow the user decide if they want sample data or not as an option, but this might increase the templates complexity. In the other hand, to generate always with sample data will make the developers to have to "delete" things they don't want. 

My opinion is that an "opt in / opt out" sample data would be great.

mrlacey:
Any issues with reusing some NorthWind data for this?

mrlacey:
@ralarcon yes, populate the SampleDatService with data from NorthWind. This should provide a varied number of options and avoid any issues with made up names, etc.

ralarcon:
Just one more question... are you thinking in just copy some of the datat into a "static" class or implement the data acces to the database? In our initial approach we where thinkin to have some static data within the SampleDataService, I don't know if adding database would make things "harder"...

mrlacey:
static data. It needs to be self-contained and easily removable.

ralarcon:
That was exacly our thougth.

mvegaca:
Pull request of SampleDataService merged to dev branch.

mrlacey:
Closing as now in dev nightly and seems to be working fine.

Issue #129, opened 2017-04-05
Page: Chart


crutkas:
Having a page with a chart would be swell.  We could use the telerik controls

mrlacey:
If happy with the default/sample data can close this out soon

mrlacey:
changing milestone as already in as part of 1.1

mrlacey:
Still to do here:
- Change to a bar chart
- add comment link to http://docs.telerik.com/windows-universal/controls/radchart/getting-started 
- update description.md with more info and link(s)

mrlacey:
Closing as now in dev nightly and seems to be working fine.

Issue #128, opened 2017-04-05
Page: Data grid


crutkas:
I would like a data rich grid.  We could use the Telerik data grid control to power this.

mrlacey:
I have a PoC of this and have shown it to Clint & Michael. Will add as a branch next week when I get back from build.

mrlacey:
There's a branch with this now at https://github.com/Microsoft/WindowsTemplateStudio/tree/Issue128-DataGrid
Feedback welcome

@crutkas this currently uses randomly generated sample names - if we need approved names (even on a community contrib item) can you please source some accordingly.

mrlacey:
changing milestone so all three new pages based on Telerik controls can go in together

mrlacey:
Still to do here:
- Change to a bar chart
- add comment link to http://docs.telerik.com/windows-universal/controls/radchart/getting-started 
- update description.md with more info and link(s)

mrlacey:
Closing as now in dev nightly and seems to be working fine.

PR #433, opened 2017-06-01
Centralized sample data

mrlacey:
#403 and X-Ref #128 & #129 

Also adds ability to have hidden features - can be dependencies but not selected on their own.


mrlacey:
@ralarcon I'll add the the existing PR

mrlacey:
@ralarcon should be good now.

Issue #249, opened 2017-05-01
Standardize code on Var or class name


crutkas:
there is a ton of spots mixing and matching var versus the actual class name.

mvegaca:
I've push 2 commits including var code normalization on madrid-dev branch.
[var code normalization](https://github.com/Microsoft/WindowsTemplateStudio/commit/7e6612b32279c0ecb56a3cab3c0cba22e411d157)
[var code review in src projects](https://github.com/Microsoft/WindowsTemplateStudio/commit/483b093975f41332e62a275bee03b9101ab905e9)

The criteria I have applied is using **var** on self-described names and I also improve name definitions.

mrlacey:
So it's **var** everywhere then.
Any movement on adding something to enforce this?

skendrot:
I love var as long as I know what the type is. As @mrlacey points out, `var input = GetInput();` isn't helpful to know what the type actually is. 

Yes we have Visual Studio and can hover over var, but I view many open source projects online through github or other sites. Seeing `var num = 0;` is fine, but not `var input = GetInput();`

ralarcon:
In my opinion, we can include StyleCop as @mrlacey suggest but we need to rationalize the style applied. We can start by defining the style rules to be applied and then pass over the code to check how far we are...  cc/ @crutkas

mrlacey:
To be clear (and as I think I've said somewhere on here before)
- I am in favor of consistent code structure and formatting, both for the code in the project and the generated code.
- If (when?) we have rules, these should be checked/enforced by tooling. If a manual step they get forgotten.
- I do not think we should include StyleCop (or similar) in the generated project.

crutkas:
Yes, Yes, Yes.  In unit tests, can we run stylecop over a generated project w/ everything checked to see if we throw anything?

ralarcon:
@crutkas @mrlacey for the code in the project we can easily enforce the validation within the build process. Build will not pass if there are violations or something similar.

For the generated project, we can explore the ability to run the tool after the generation and check if there are some violations.

It would be easy to enable the usage of VS Code Analysis with set of rules we like. The execution and reporting would be simple. Anyway I understand StyleCop contains a broad number of coding style rules. Does it make sense to just enforce thru VS Code Analysis?

Passing to action:
@mrlacey it would be helpful if you can aid running the StyleCop tool against the project code and generated code to see how far we are. The sooner we include the enforcement, the better, but we need to check the impact.

[...]

PR #451, opened 2017-06-13
Issue249 stylecop

mrlacey:
Raising PR now to make it easier to discuss progress on StyleCop integration #249


ralarcon:
@mrlacey changing the target brach to DEV. Just reviewing this...

Issue #364, opened 2017-05-18
Hamburger menu generated by template does not match typical UWP style.


geometrikal:
- Row height is not 48 x 48
- Animation is wrong

For the row height etc, the following works ok, but not sure how to change the animation.

    <Button
                Margin="0"
                Padding="0,0"
                HorizontalAlignment="Stretch"
                HorizontalContentAlignment="Stretch"
                AutomationProperties.AccessibilityView="Control"
                AutomationProperties.LandmarkType="Navigation"
                AutomationProperties.Name="{x:Bind Label}"
                Background="Transparent"
                BorderThickness="0"
                Click="NavigationButton_Click">
                <Grid>
                    <Grid.RowDefinitions>
                        <RowDefinition Height="48"/>
                    </Grid.RowDefinitions>
                    <Grid.ColumnDefinitions>
                        <ColumnDefinition Width="48" />
                        <ColumnDefinition Width="*" />
                    </Grid.ColumnDefinitions>
                    <Rectangle
                        Width="6"
                        Height="36"
                        HorizontalAlignment="Left"
                        VerticalAlignment="Center"
                        Fill="{ThemeResource SystemControlForegroundAccentBrush}"
                        Visibility="{x:Bind SelectedVis, Mode=OneWay}" />
                    <FontIcon
                        HorizontalAlignment="Center"
                        VerticalAlignment="Center"
                        FontSize="16"
                        Foreground="{x:Bind SelectedForeground, Mode=OneWay}"
                        Glyph="{x:Bind SymbolAsChar}" />
                    <TextBlock
                        Grid.Column="1"
                        Margin="0,0,24,0"
                        VerticalAlignment="Center"
                        Foreground="{x:Bind SelectedForeground, Mode=OneWay}"
                        Text="{x:Bind Label}"
                        TextTrimming="CharacterEllipsis" />
                </Grid>
            </Button>

mrlacey:
For note, differences are:
- Removed button padding
- Set Grid row height = 48
- Set Rectangle height = 36
- Set Rectangle alignment = Center

mrlacey:
Please note that the ShellPage will use the HamburgerMenu from the community toolkit in a forthcoming update and addresses most of these already.
The only thing it doesn't do is the rectangle size change. @geometrikal what's your source for the rectangle height change

mrlacey:
@geometrikal what's wrong with the animation?

geometrikal:
[...]

PR #452, opened 2017-06-13
HamburgerMenu styling

mrlacey:
HamburgerMenu now matching Fall Creators Update redlines
#364


crutkas:
Any reason why we should **not** we migrate this also to the UWP Community Toolkit?

mrlacey:
@crutkas The Toolkit appears to have bigger plans for converting the HamburgerMenu to include additional functionality that exists in the NavigationView control and not just updating the styling as I have done. What we now have may help them but won't be a complete solution for their needs.

X-Ref https://github.com/Microsoft/UWPCommunityToolkit/issues/1185

ralarcon:
@mrlacey @crutkas a cuple of questions:

Is it worthy to have a dependency to the toolkit and override the styles to improve the visualization in our apps but still depend on the toolkit? 

Would this block the users to take advantage of future improvements in the UWP Toolkit control? Why do not move this to the toolkit even they have bigger plans?

Apart from this, I would suggest to move the new style defined to a Style sheet within the Styles folder.

crutkas:
I think we should do it short term here, have a patch over for them.  When they update with updated styling, we remove with template hot fix.  @nmetulev do you agree?

skendrot:
See my [latest comment](https://github.com/Microsoft/UWPCommunityToolkit/issues/1185#issuecomment-309096605) on the issue.
If anyone here knows an easier way to solve this I'd love to hear it

crutkas:
noted a title gap (also grid doesn't have a title too)
chartbug
titlediff

mrlacey:
@ralarcon not sure what you mean by this

> Why do not move this to the toolkit even they have bigger plans?



@ralarcon I'd not put the style into a separate file as it's only used in this location and there should be no need to have multiple top level navigation controls. Splitting it up would just add a slight delay to load time and separate two things which should always be together,

@crutkas the 8px gap you call out is in the redlines. If alignment with content on the page is the issue that's because the pages don't match the redline guides for size, alignment, and positioning. I've raised this a separate issue (#472) but you should make a call if this is necessary for the 1.1 deadline. (It's just in the templates so could change after the VSIX release.)

ralarcon:
@mrlacey I was trying to reflect that I think this changes should be included in the UwpCommunityToolkit even if they are going to change / modify other things for RS3. 

Approving the PR...

mrlacey:
yes, the code is there if the toolkit want to use it but there no evidence that they do and they, they seem to have other, larger plans, and have deliberately gone a different way in their styling of that control. I'm not going to force this on them

skendrot:
@mrlacey What do you mean by:
> but there no evidence that they do
I have been trying to get the styles into the core control as noted in this PR already. If you have a solution that works please feel free to submit a PR

mrlacey:
@skendrot What I meant is that the toolkit isn't trying to make the same change I've made here. The discussion seems to be about mimicking the functionality of the NavigationView as much as the visual styling. We only do the styling.  
We can also do other things more easily as we're a single specific style and not be needing to cope with some of the issues that come from being so general. For instance, Because we have a single visual style it was easy to ensure a single ListItem template has the styling for the selection indicator and sets size and spacing for the icon and text. 
The way we track which item is selected is also specific to our implementation and not done in the way they'd need to which is more generic.  

If they want to use what I've done for reference then that's fine but I don't know what it would provide them that they can use directly apart from any specific size or spacing figures but I'd assume that they have those anyway.

nmetulev:
I looked through your styles, and we could probably use the changes you've done to the default template. Of course the item template changes are specific to your implementation and would not be as straightforward to bring in. 

If you are ok with bringing in just the control template changes to the toolkit, let me know and I can help. They should cut down a bit of your custom style.

nmetulev:
I created a new branch on the toolkit side with those changes already: https://github.com/Microsoft/UWPCommunityToolkit/tree/hamburger

crutkas:
@nmetulev it is a starting point :)

nmetulev:
Created PR here: https://github.com/Microsoft/UWPCommunityToolkit/pull/1229

nmetulev:
The PR was merged in the toolkit and will be available in the 1.5 release

skendrot:
Final NavView styles into UWP Toolkit in [PR #1239](https://github.com/Microsoft/UWPCommunityToolkit/pull/1239)

PR #463, opened 2017-06-16
update docs

crutkas:
fix extension info


crutkas:
top part looks to have been merged a while back, only bottom is relevant.  small verbiage change

Issue #321, opened 2017-05-12
Enable the creation of Xamarin.Forms apps


mrlacey:
From #319 
> Is there a plan to also include templates to help kick-start Xamarin and Xamarin.Forms development? Thinking along the same lines, wouldn't it be great to have the ability to create a Xamarin Forms solution, pick which MVVM framework to use, choose between a hamburger menu or a tabs based app, include settings, themes (light-dark), add maps, login page, splash screen etc. Essentially everything that the Template Studio offers for UWP apps.

Please add any specifics about what you'd like to see for Xamarin.Forms below.

https://github.com/Microsoft/WindowsTemplateStudio/wiki/Xamarin

crutkas:
how is this different than #229?

mrlacey:
#229 is to create a solution containing a UWP app project and a library project (PCL or Net Standard) that could then be used in a Xamarin.Forms (or other) project. It doesn't create anything that targets another platform though.
This issue is a request to be able to create a Xamarin.Forms app from the start. The output from this would be a shared library (or PCL) and multiple platform heads (iOS, Android, Windows, & others?) as if doing File>New>Xamarin.Forms. 
This is the request for the ability to do for Xamarin.Forms, what WTS does for UWP.

mrlacey:
It appears something might be happening in this area....https://github.com/Microsoft/WindowsTemplateStudio/tree/xamarin

crutkas:
Added wiki page

nguyenthanhliemfc:
Is it support Xamarin-forms now?

crutkas:
We have multiple aspects of this built out, we are forming a plan of exact execution.  We'll be sharing shortly for feedback

nguyenthanhliemfc:
@crutkas Thank you

mrlacey:
@nguyenthanhliemfc what, specifically, do you want from Xamaring.Forms support? What should (and shouldn't?) it include or enable you to do?

LocalJoost:
I would use this, and this would definitely help us. We now suffer a lot from 'set up costs' and it makes apps too expensive compared to web apps.

jesulink2514:
@mrlacey I think that bring Prism for Xamarin.Forms templates could be a great start point considering that Prism Template Pack already exists. https://marketplace.visualstudio.com/items?itemName=BrianLagunas.PrismTemplatePack

anotherlab:
[...]

PR #478, opened 2017-06-19
Include  "Universal Windows" in VS Project Template name

ralarcon:
The Project Template name is replaced at build time in order to change it for each environment (loca, dev-nightly, pre-release and pro).

Updated the name in the local template and changed the builds variables to include "Universal Windows" as part of the name.

### DIFF
--- a/code/src/ProjectTemplates/CSharp.UWP.2017.Solution/CSharp.UWP.VS2017.Solution.vstemplate
+++ b/code/src/ProjectTemplates/CSharp.UWP.2017.Solution/CSharp.UWP.VS2017.Solution.vstemplate
@@ -3,7 +3,7 @@
             Type="ProjectGroup" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005" 
             xmlns:sdk="http://schemas.microsoft.com/developer/vstemplate-sdkextension/2010">
   <TemplateData>
-    <Name>Windows Template Studio (local)</Name>
+    <Name>Windows Template Studio (Universal Windows; local)</Name>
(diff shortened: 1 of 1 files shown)


crutkas:
how will this work in once we start supporting xamarin?

mrlacey:
It depends on how you want the Xamarin integration to work.

If we think about it on its own then it makes sense for an option to create Xamarin projects to be under the Cross-Platform group (see image below). In this case, it would make sense to have the bracketed suffixes to distinguish between multiple items in the single list of all templates.

image

If the intention is to add Xamarin[.Forms] as another option under frameworks in the current wizard, it might be weird that getting to this is via a template in the "Windows Universal" folder. It would be fine for people wanting to create a UWP app with XF ready to move to other platforms but wouldn't be great for people specifically wanting to target multiple platforms from the start.

I expect that Xamarin would prefer to have a separate template entry. I think it's the better option too, but it does raise some questions:
- Should you also be able to select Xamarin as a Framework for the UWP version?
- If targeting Xamarin.Forms as a framework, what about using other frameworks (MVVM Light, MVX, etc.) with it? as this is popular.
- If there is a cross-platform entry should that only be limited to creating Xamarin projects?
- Would both templates be installed together? Even if the machine only has one of the UWP or Xamarin tooling installed?
- and more questions for another time.... ;)

crutkas:
Most of these should be in the XF issue.

- I think you'd be able to see Xamarin in UWP area
- the framework question comes into massive play here as a best POR.  I think the wizard will need to be tweaked no matter what.
- dealing with tool detection is a great item to bring up.  I didn't think about that

crutkas:
xref #321 for XF issue

ralarcon:
I fill more natural to have an entry under "cross platform" for xamarin; this would just run the wizard pre-filtered because I guess xamarin would have different frameworks and project types too...

Issue #477, opened 2017-06-19
Feature: Ability to include standard text at the top of all files


mrlacey:
It's common to want a specific piece of text at the top of each file (of certain file types) often as a copyright header and it would be nice if this could be added by just adding a feature during project generation.

Some questions:
- How best to specify what the contents of the header should be in the wizard?
- Which file types to add this to? Just *.cs, or *.xaml as well?

ralarcon:
Initially I was thinking the scope of this limited to the code that we generate including our license disclaimer but, if i'm not wrong, this is something you want to include as an end-user wts feature in the wizard...   

We can start using this feature as a paradigm and discovery for features which requires extra user inputs. Good chance to start this area.

mrlacey:
Also related to the idea of "enhanced pages" #325

crutkas:
idea here is a user wants this out of the gate for their solution?  clever.

mrlacey:
Yes, the idea here is that if I'm using this to build an app I want copyright notice and comment headers in all files. I want these included in the generated project, not to have to go and add to all the files that are created.

crutkas:
Great idea. Jelly I didn't think of it :)

PR #481, opened 2017-06-19
bugfix for sort-usings exception

KarlErickson:
Small fix for an exception that appears only when a template file has content before its using statements (e.g. comments). In that case, LastIndexOfWhile can return a value that is less than startIndex, causing problems in the caller. (For existing templates, startIndex is always 0, so this problem doesn't arise).


mrlacey:
Have you raised a fix for something that doesn't exist in any of the templates?
I don't expect us to ever include something before using statements in default templates.

It would be good to also see tests to cover a potential change in functionality.

KarlErickson:
Added a test.

I'm creating in-house templates for samples that include license boilerplate before the using statements in every file. This doesn't affect built-in templates, but it would affect any user who wanted to do something similar in their own usage of the tool. 

Not sure if e.g. #477 work would be broken without this, but the current LastIndexOfWhile implementation does not match its implied semantics.

crutkas:
looks fine to me.

Issue #509, opened 2017-06-23
Bug: Broken links in page descriptions


mrlacey:
There are broken markdown links for the three new pages added in 1.1.

Image below shows MediaPlayer but Chart & Grid are similarly affected.

![image](https://user-images.githubusercontent.com/189547/27497717-fa094a90-5853-11e7-98b6-34174f8f8cc0.png)

Grid description also mentions chart in a horrible display of a cut-and-paste schoolboy error.

It turns out that some markdown viewers are more tolerant of this than our code to display these files.

crutkas:
@mrlacey feel like we can close this out.

PR #510, opened 2017-06-23
Fix markdown description links

mrlacey:
#509

Submitting this against /dev but would be good to get this fix out in an updated template ASAP, rather than wait for the next VSIX update.


crutkas:
i think this may need to be a hot fix.  @ralarcon do you agree?

Not merging until we figure that out.

Issue #295, opened 2017-05-11
Add out of box / first run message


mrlacey:
Add a way to show a specific message to users when using the app for the first time only.

mrlacey:
I have a PoC of this and have shown it to Clint & Michael. Will add as a branch next week when I get back from build.

mrlacey:
Confirmed in DevNightly build with wizard version 0.14.17186.65008 and template version 0.14.17186.7

PR #523, opened 2017-06-27
Add OOBE message prompt

mrlacey:
#295


mrlacey:
@ralarcon suggested change made

Issue #296, opened 2017-05-11
Feature: display message after app update


mrlacey:
Add feature to detect when installed app has been updated and show a message to user listing features, bug fixes, etc.

mrlacey:
I have a PoC of this and have shown it to Clint & Michael. Will add as a branch next week when I get back from build.

mrlacey:
Confirmed in DevNightly build with wizard version 0.14.17186.65008 and template version 0.14.17186.7

PR #524, opened 2017-06-27
Display dialog after app update

mrlacey:
#296


mrlacey:
@ralarcon suggested change made. Good catch

PR #548, opened 2017-06-30
Allow for custom AI Endpoint in config file + tagged AI key GUID

lucastaiMSFT:
Provides for optional custom telemetry endpoint in config file
Tagged AI iKeys won't fail parsing and report as unavailable or invalid


crutkas:
Synced with Lucas on this.  We're just using a custom Visual Studio endpoint but making it so anyone can use App insights as well.

Issue #461, opened 2017-06-15
Ensure our generated code meets our stylecop bar


crutkas:
We want to be sure the template code also meets our style cop quality bar.  This should not include a style cop integration for the final template, but should happen during our testing and know that we are producing code that will not throw errors.

ralarcon:
@crutkas @mrlacey We have in mind a solution approach for this. We can implement a feature which includes the style cop nuget reference + the global supressions we consider + the build configuration to enforce it.

The feature could be hidden if we want only add it in our build process or even, provide it publicly to allow users to take advance of it...

What dou you think?

crutkas:
we can add it as an option to expose to public.  

I view this low priority work.  My opinion is the code analysis against only the all options enabled generated projects because it does add some time in build process.

Any reason why we wouldn't use our suppression file?

ralarcon:
We keep in mind this to implement it when possible and check how impacts in the build (maybe we can only add stylecop checking to the projects which includes all the features, that is, 9 projects. 

We can use our suppression file, of course ;-)

mrlacey:
We shouldn't use the same suppression file as we use in the tool code because:
- it includes exclusions which are currently only temporary and we will eventually (hopefully) remove
- it doesn't exclude things which don't apply to generated code. (e.g. file copyright headers)
- it doesn't correspond with [the guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) we claim to follow.
- there are issues we previously excluded but have now been addressed but still exist in the generated code.
- even with our current suppression set, there are inconsistencies that are highlighted in the generated code.

Rather than vague theorizing about this, I feel it just needs to be done.

ralarcon:
Agree. We can setup the suppresion file we want. My understanding is that as time goes, we will be improving our supression to reflect all things you claim, so we can use the same quality bar for generated code and for our code. It is not possible to go 0 - 100 in one shot.

By the way, this is not theorization, it is more a solution aproach to be implemented in short term.

mrlacey:
Have created a branch at https://github.com/Microsoft/WindowsTemplateStudio/tree/Issue461-StyleCopOfGenerated    
Will hold off merging with dev until after 1.1 is released.  
Points of note:
- Had to disable tests running in parallel for this to work (also have done this in the VB branch.) The generation code is very tightly coupled to some global static objects which make testing tricky. I've created #497 to track improving this.
[...]

Issue #497, opened 2017-06-21
Improve speed and isolation of tests


mrlacey:
Would be good to improve the code isolation so that tests can be run in parallel and can, therefore, be executed faster.

mrlacey:
Be good to wait until after the VB and StyleCop branches are integrated before attempting anything here or risk making merging them *very* tricky.

crutkas:
what is not making the tests run not parallel?  looking at it, each test is 1 after another versus multiple at the same time

mrlacey:
> what is not making the tests run not parallel? looking at it, each test is 1 after another versus multiple at the same time 

A consequence of the VB changes (plus they don't run in parallel on the build server)
There were also intermittent errors previously when running some tests in parallel - e.g. #709

The generation tests can't run in parallel as they are based on the templates appropriate to the test. Under the hood, there's a lot of copying of files to specific folders based on the selected items (project type, framework, pages, features, and now language.) Tests can't run in parallel with this or they start overwriting the files that different tests need. Imagine one test starts and copies over some C# files and then another test running at the same time overwrites these with some VB ones. The current code it too brittle to handle such testing.

I raised this issue directly in response to realizing how things currently work and how this lack of isolation prevent parallelization.

A short, quick fix would be to have each test put files in different folders but I haven't verified: this doesn't have other consequences; what this means for non-test generation; what the required code changes are; or if it just ignores another underlying problem in the way the generation code is structured. (The big static globals being a red flag on this too.)

mrlacey:
Past changes to cache template unpacking by tests have made running tests much faster. This makes this redundant

PR #559, opened 2017-07-03
Run StyleCop on all generated projects

mrlacey:
#461
Now run StyleCop on a generated project containing all pages and features


mrlacey:
1. **In terms of adding StyleCop as a feature:**

By all means, create a new issue to track this but be aware of the following:
- We'll need to test all combinations of projects, frameworks, pages, and features against the rules we provide. It would be really bad to have this as an option but our generated code doesn't pass the checks. (I strongly suspect that there are combinations of pages/features which leave blank lines, etc. but this is not observed when including them all.) These tests will have a significant impact on overall test time which will be an issue until we improve test performance. (see below for more on this.)
- What suppressions, if any, should we include by default? What makes sense for a general use case? It's definitely not what we use in the core code. It may be the suppression rules in the tests I've just added. It's not enough to say we should enforce a standard we have to be clear about what that standard is and why.
- As per the original issue. having it is a low priority. Would also be good to get some wider feedback to see if this is something that's wanted before we spend time on it. ...
- In my experience code consistency isn't something that's widely enforced and lots of developers have strong negative feelings towards it. I've pushed this for WTS because I think it's important. It's also important to me that the code that's generated is of a high standard and formatted consistently. This is why I've spent so much time on it. My attitude to this is not common among other developers.

What do you mean by this?

> To be able to execute and make the test fail, it is required to create a custom build configuration and use it from the build batch (we need to figure out how to exactly do it).

The above code modifies the existing build config as part of the test. No need to create a custom build configuration.

2. **Parallelization**

This discussion should probably happen as part of #497 (Improve speed and isolation of tests) which I created when I first identified the issue of tests parallelization.  
The big issue was (I haven't fully checked if this is still the case since the right-click merge) that all the templates in use for a specific test were copied to a single folder where they were extracted from the zip. When different tests were running at the same time, multiple things were being deleted, copied, and unzipped at the same time forcing tests to fail in strange ways as files would be locked, disappear, or appear at the wrong time. Fixture (trait) isolation doesn't solve this because different fixtures use different combinations of templates. (This caused lots of issues with the VB integration when the templates were changed mid-test and suddenly it was adding files in a different language!)
Yes, this needs to be reviewed. It's why I created #497.  
I've also noticed that individual tests are taking longer to run since the right-click changes but haven't had the time to dig into it as thoroughly as it needs.  
There are definitely issues with the reuse and reloading of templates in the current code that not only affects parallel testing but also the speed of individual tests.

ralarcon:
Answering to your points:
* We have the test which add all the features. If this test fail, then we need to find the "page/feature" responsible.
* We can have the feature hidden by now, with the suppresion file we need to be able to progressively remove all the warings / errors.
* Once our generated code pass the StyleCop rules, we can make the feature public and release it.
* You are right regarding the build configuration. Your approach is better and simple.
* Regarding parallelization: I know that the issue becomes from the fact of sharing folders. I just need some time to improve the "LocalTemplateSource" to be able to use multiple targets folders in case we need... 
* Regarding time spent: now we have more templates and the templates are being packaged and unpackaged so it would take longer. As a workarround for you, you can run locally just the trait "ProjectGeneration" and let the build server to run the one-by-one.

If you agree, 
* You can reuse all the StyleCop stuff you have implemented to create a new "first citizen" feature template using the hidden property.
* Include the suppresion file which is required to be able to progress.
* The parallelization can be followed in #497 with the v1.3 as target.

mrlacey:
> We have the test which add all the features. If this test fail, then we need to find the "page/feature" responsible.

Not sure if this is a question or a statement. If a test fails then the test output will detail the error from the test. This includes file name(s) and line number(s). Post action merging means that might not be where it is in the template but any failures will be in the files modified or added as part of the change. The project created as part of the test is also kept so that can be used as a reference if needed to find the cause of the failure.

> We can have the feature hidden by now, with the suppresion file we need to be able to progressively remove all the warings / errors.

I'm not sure what you mean by this. Are you talking about adding a public template for including StyleCop in a generated project? If so then this is a separate task. Raise a new issue for it. I also don't think many people will want it. Please can we wait for people to ask for it before spending any more time on it.

Not sure what you mean by being "able to progressively remove warnings or errors." Do you mean remove entries from the suppression file? I've included justifications for all the entries that are there but can make a stronger case for any/all of them if you think there are ones that should be removed.



> Once our generated code pass the StyleCop rules, we can make the feature public and release it.

If adding a new feature it will get picked up in the automated testing. If the new template code and changes don't pass the StyleCop test then the tests will fail and it shouldn't be checked in. It should NOT be a case that anyone adds a new template and checks it into, or merges with, dev if it isn't ready to ship. That means being of the standard we require.

> Regarding parallelization: I know that the issue becomes from the fact of sharing folders. I just need some time to improve the "LocalTemplateSource" to be able to use multiple targets folders in case we need...

Yes, do this as part of #497

> Regarding time spent: now we have more templates and the templates are being packaged and unpackaged so it would take longer. As a workarround for you, you can run locally just the trait "ProjectGeneration" and let the build server to run the one-by-one.

Not sure what you mean by this. I don't need a workaround as able to run individual tests. I'm looking to #497 to address the overall time for running all tests.

> You can reuse all the StyleCop stuff you have implemented to create a new "first citizen" feature template using the hidden property.

As mentioned above, gather feedback on whether people really want this before we spend any more time on it.

> Include the suppresion file which is required to be able to progress.

Include it where? It's at https://github.com/Microsoft/WindowsTemplateStudio/blob/Issue461-StyleCopOfGenerated/code/test/Templates.Test/TestData/StyleCop/GlobalSuppressions.cs

> The parallelization can be followed in #497 with the v1.3 as target.

Yes, a separate issue should be tracked separately. Do not hold this up for a separate issue. Prioritization can be discussed separately.

ralarcon:
@mrlacey All your comments are valuable, and it is probably that my explanations are not clear enough. I'm just looking for synergies.

What I originally proposed in #461 was that an option to address the requirement to run stylecop against the generated code was to create a feature to be included in the tests. 

Using the same effort (more or less) we can release another feature to be used by anyone who want to do so. By the way, sometimes we need to make implementations that are not being asked by the users / community... it would depend on other engineering targets or requirements.

Regarding the suppression file I mean just the opposite (if required): to include more suppressions to do not block the progress and start removing those added progressively.

Regarding the ability to detect if a developed feature is compliant against the Style rules, how we can enforce it apart from running in the tests? This is what we do right know. Someone could push and merge code that is not compliant with the style rules and it will be detected in the build... what is your approach?, probably I missed something.

Regarding the workaround,  I just was trying to help. I know it is anoying to wait for the test to finish.

What are the benefits injecting / copying some code on the fly while running the tests vs doing it with a feature considering the test infrastructure already allow it, that by the way using make our system stronger...

mrlacey:
> What I originally proposed in #461 was that an option to address the requirement to run stylecop against the generated code was to create a feature to be included in the tests.

That's what I've done, just not made the feature public. The feature is only included in the test. The ability to do this with tests also presents some potentially interesting options for future testing too.

> Using the same effort (more or less) we can release another feature to be used by anyone who want to do so. By the way, sometimes we need to make implementations that are not being asked by the users / community... it would depend on other engineering targets or requirements.

Doing something just because you can is a weak argument.  
Yes, we can do things that aren't being asked for. But that doesn't mean that you should ignore what people are asking for. Effort spent onn this is effort not spent on things we know people want. There is no engineering target or requirement to make this publicly available.

> Regarding the suppression file I mean just the opposite (if required): to include more suppressions to do not block the progress and start removing those added progressively.

To add more rule suppressions seems like an argument against consistency. Without specifics of what you want to start excluding it's hard to discuss this thoroughly but it's a slippery slope to just start adding suppressions because someone feels it makes things easier for them.   

> Regarding the ability to detect if a developed feature is compliant against the Style rules, how we can enforce it apart from running in the tests? This is what we do right know. Someone could push and merge code that is not compliant with the style rules and it will be detected in the build... what is your approach?, probably I missed something.

The tests should be a check. **No one should be pushing and merging code that hasn't been tested.** So that's not a valid argument.  
Knowing what the code should look like in terms of structure and formatting (the biggest StyleCop issues) shouldn't be an unknown. An awareness of this when creating templates shouldn't be a big issue and is something I'd expect of the most junior developer. With this knowledge, creating compliant templates is not difficult. If anyone is having specific issues it'd be good to understand what those are.

> Regarding the workaround, I just was trying to help. I know it is anoying to wait for the test to finish.

Yep, let's focus some effort on making all the test faster #497 

> What are the benefits injecting / copying some code on the fly while running the tests vs doing it with a feature considering the test infrastructure already allow it, that by the way using make our system stronger...

I have the feature just in tests because:
- the original requirement was just to do this in testing
- there was no requirement for making this a public feature
- The existing infrastructure doesn't support having features that aren't public
- The ability to hide features (like with the SampleData source) also excludes them from testing (they only get added as dependencies)
- It didn't make sense to make a second kind of "hidden" feature which should be with the other templates but was only used in testing. Adding the logic for this would just further complicate the logic of templates selection that already exists. Having it separate keeps that logic separate too. I don't believe that adding complexity makes the "system stronger".


If you really want to add it then go ahead.

PR #608, opened 2017-07-05
Localization infrastructure

ralarcon:


mrlacey:
Is there an issue discussing the changes here and the reasoning behind them all.  
I'd also like to understand the implications for existing and future templates on these changes.

crutkas:
do we need to localize output / logs?

@mrlacey  templates *should* all localized by default.  This should be for the wizard only.  What file makes you think that?

crutkas:
If a json or md file isn’t localized, it defaults to English, correct?

sibille:
Yes, if the file is not there it defaults to english

crutkas:
if that is the case, why did we check in all the languages versus only the ones we wanted.  Now we have to update every variant if a language adjustment is needed right now.

PR #629, opened 2017-07-06
Fix padding on changes summary licenses

mvegaca:


crutkas:
I feel like this is a bad fix here.  You're adding another item on the visual tree just to fix a padding issue?  Why not just properly correct the margin or correct the margin at the parent level?

Issue #618, opened 2017-07-05
redundant information w/ screen VS info modal.


crutkas:
![image](https://user-images.githubusercontent.com/1462282/27886274-33e6dd96-6190-11e7-9964-11b3708bcadb.png)

This is related to #596 where we display redundant information.

mvegaca:
Updated on following [pull request](https://github.com/Microsoft/WindowsTemplateStudio/pull/630)

- Replace redundant loading text by Windows Template Studio Logo
- Add LogoControl with customizable fill color property

![image](https://user-images.githubusercontent.com/5556784/27900793-eea3645c-622f-11e7-813e-687d554d5350.png)

ralarcon:
This will be soon fixed over the existing control instead of changing the control type.

mvegaca:
Fixed on #677

mvegaca:
Checked on:
Wizard Version - 0.14.17194.65002 
Templates Version - 0.14.17194.2

Issue fixed and closed.

Issue #570, opened 2017-07-03
user control vs control


crutkas:
why is it togglepane.xaml and togglepane.cs.  shouldn't this be togglepane.xaml.cs?  

This also affects overlaybox and codeviewer

mvegaca:
User control combines one single Xaml file with the cs file, for example on Status control.
I used prefer to add Custom Control with separated Xaml and CSharp code files because I prefer the use TemplateBinding when I make UI controls.

crutkas:
lets pick one and move forward.  

Unless there is a clear reason, we shouldn't have multiple competing ways of doing things.  Right now, other than developer preference, there was no reason to do something different.

crutkas:
feels like we're migrating pure control instead of user control, correct?

At this point, i think we can punt this 1.4.  Low hanging fruit for just general app cleaniness

mvegaca:
I agree with you

mvegaca:
Updated from Control to UserControl in #1125. From now, we will use UserControl.

dgomezc:
Verified in dev-nightly:
Wizard: 0.14.17284.65001
Templates: 0.14.17284.1

PR #630, opened 2017-07-06
Update loading feedback

mvegaca:


crutkas:
Why are you changing control types?  This feels like change just because of preference?  This goes back to #570.  There isn't a clear reason to change this.

mvegaca:
I prefer to use Control instead of TemplatedControl because it offers more flexibility and the possibility to use TemplateBinding to access control dependency properties.

crutkas:
@mvegaca that is developer preference and a design shift that should have been discussed.  Template binding versus just direct binding, no clear reason to adjust and break an existing pattern for a single control.

If we were going to do this, we should have had a nice quick discussion, and converted all the controls, not just one-offs.  same stuff 100% with template generated code.  we really should be living and breathing best practices.  

Right now, that isn't something that we should be doing for 1.2 release as well, we should be on stabilization and bug fixes.

ralarcon:
Closing this PR. #618 will be fixed with the same approach using the existing control.

ralarcon:
#570 is the issue for discussion.

PR #631, opened 2017-07-06
Fix clickable area on changes summary list items

mvegaca:


crutkas:
why is this triple grid nested with a fixed width?  Feels like that control can be simplified.

PR #638, opened 2017-07-06
Avoid concurrrent contents updates

ralarcon:


crutkas:
walk me through how this is going to fix the issue?  To me it seemed as if the files were still in use.  What was changed that would have caused this to appear now so much versus before?

ralarcon:
The issue become from the fact that having a syncronization on the fly, another one start doing the same actions (this is even worst in the Local instance as the targets are the same). The sequence is more or less:

-> Donwload content to a certain temp folder 
-> Take note of the temp folder used
-> Uncompress the content to other temp folder and copy
-> Take note of the temp folder used
-> move content
-> Cleanup temp folders used

If the syncronization occurs twice consecutive, the while the second thread si uncompressing content the first process is trying to cleanup the content.

Avoiding two syncronizations concurrently remove the collision. Because of this I locked the Sync and the Check for update methods using the LockSync() and UnlockSync() (which uses a critical region based on a flag var). I tested it and works fine.

The fist commit contains other I used during the investigation which I pushed by error, but I undoed it.

crutkas:
But this also doesn't solve the issue for silently failing.  If i didn't notice the output log, i wouldn't have been able to figure out what was happening.

ralarcon:
Yes, that is true. We can make it fail or do not clean up the temp folders and remove those from time to time (like the temp generation folders).

PR #639, opened 2017-07-06
Double check open temp folder availability

ralarcon:


ralarcon:
Branch selected erronously

PR #642, opened 2017-07-06
Update url and increase autohide seconds

mvegaca:


crutkas:
why was the timer changed?

ralarcon:
To leave the auto-hide messages a bit longer time. In some connections speeds we have detected that the time between donwload the content and being able to extract and refresh the templates takes a while so there is an opportunity to have no feedback message, more over if we finally replace the "Loading..." by an image in the main window as proposed in #630.

Issue #648, opened 2017-07-06
Local crash on right click add in emulator


crutkas:
On either "+" button i click, it crashes

public override ValidationResult Validate(string suggestedName)
        {
            var existing = Directory.EnumerateDirectories(_config)
                                            .Select(d => new DirectoryInfo(d).Name)
                                            .ToList();

System.IO.DirectoryNotFoundException: 'Could not find a part of the path 'C:\Users\crutkas\AppData\Local\Temp\WTSTempGeneration'.'

System.IO.DirectoryNotFoundException occurred
  HResult=0x80070003
  Message=Could not find a part of the path 'C:\Users\crutkas\AppData\Local\Temp\WTSTempGeneration'.
  Source=<Cannot evaluate the exception source>
  StackTrace:
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileSystemEnumerableIterator`1.CommonInit()
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.Directory.EnumerateDirectories(String path)
   at Microsoft.Templates.Core.DirectoryExistsValidator.Validate(String suggestedName) in C:\Users\crutkas\Source\Repos\NewRepo\code\src\Core\Naming\DirectoryExistsValidator.cs:line 27
   at Microsoft.Templates.Core.Naming.<>c__DisplayClass2_0.<Infer>b__0(Validator v) in C:\Users\crutkas\Source\Repos\NewRepo\code\src\Core\Naming\Naming.cs:line 32
   at System.Linq.Enumerable.All[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.Templates.Core.Naming.Infer(String suggestedName, IEnumerable`1 validators, String inferWith) in C:\Users\crutkas\Source\Repos\NewRepo\code\src\Core\Naming\Naming.cs:line 32
   at Microsoft.Templates.VsEmulator.Main.MainViewModel.GetTempGenerationPath() in C:\Users\crutkas\Source\Repos\NewRepo\code\test\VsEmulator\Main\MainViewModel.cs:line 226
[...]

PR #651, opened 2017-07-07
Ensure Temp Directory is there

sibille:
Fixes bug #648


crutkas:
what caused this new behavior to happen

PR #664, opened 2017-07-07
correct a typo about "VS 2107"

arcsinw:


crutkas:
already adjusted in dev,

Issue #674, opened 2017-07-09
Use 24 hour format consistently in output


mrlacey:
All datetimes formatted for output should use hours in the 24 hour format

mrlacey:
implemented as part of the fix for #692

PR #675, opened 2017-07-09
format datetime to use 24 hour format hours

mrlacey:
#674


crutkas:
should we think about creating a common class to pull out critical info like datetime so this doesn't happen again?

mrlacey:
> should we think about creating a common class to pull out critical info like datetime so this doesn't happen again? 

Add it to the general tidying up list.

crutkas:
https://github.com/Microsoft/WindowsTemplateStudio/wiki/Tidy-up-list

PR #680, opened 2017-07-10
Fix licenses terms links

mvegaca:


crutkas:
what happened here?  the PR fixes seem orthogonal to the bug.

mvegaca:
I'd fix a little problem with a control ZIndex

PR #686, opened 2017-07-10
Fix same file read on code viewer

mvegaca:


crutkas:
what bug does this fix?

Issue #696, opened 2017-07-10
Grammar issues in some string resources


mrlacey:
Some of the strings added for 1.2 need some tweaking from a native English speaker. ;)

Some logistics questions:
- what's with the 's' in the comments fields of some RESX entries?
- What's the current situation with localization? Do I have to copy the changes to every language version? If the is, is there a script to do this automatically so I can avoid any cut and paste issues?

mrlacey:
Branch created with changes only made to the default language files. 

Branch: https://github.com/Microsoft/WindowsTemplateStudio/tree/Issue696-GrammarUpdates

PR not raised as unsure about what to do for localized files.

crutkas:
i like it.  i need to review all the text as well still.  Blocking off time tomorrow

ralarcon:
Do not get worried by the language variation files for any of the localized resources. We have a small tool thought to be run before sent to to localization extracting (by now) all the variations from the english / default resources.

Make the PR and once we merge it, we will launch the tool to generate all the varitions properly.

mrlacey:
reassigning to @crutkas as a reminder to review the text

mrlacey:
@ralarcon where is this "small tool"? It should be part of the repo if needed for the project.

ralarcon:
Yes, it is part of the repo: https://github.com/Microsoft/WindowsTemplateStudio/tree/dev/code/tools/Localization

By the way, removing rightnow the "_tools" folder as it is not used.

mrlacey:
Setting the milestone as a reminder this needs revisiting before 1.3 is released.

Some of the file descriptions used when doing right-click>Add could use some tweaking too.

crutkas:
@mrlacey, did this get fixed?

mrlacey:
@crutkas It doesn't look like it. This could probably just be retargeted for m1.4 as it needs doing before everything goes out for localization.

mrlacey:
@crutkas PING. move this to 1.5 or close? Should really be done (if not already done) before getting translations.

crutkas:
we've already sent it to localization

ralarcon:
@crutkas did you have a chance to review this?

ralarcon:
No issues reported... can we close this?

PR #719, opened 2017-07-11
Remove _tools folder not used anymore

ralarcon:


ralarcon:
As mentioned in #696

crutkas:
if we removed this folder, we should remove it from solution files as well?

ralarcon:
Yes. I'll do.

Issue #741, opened 2017-07-11
Merge MD summary should be a proper template in Resx


crutkas:
Right now it is too hard to alter the holistic text from the completed MD file since it is 100% created in code behind.  I feel like this should have a larger resource template that has the needed parts auto generated in.

ralarcon:
We evaluate that but as the MD is quite simple to generate we did it in code... one advantage of using a template is that maybe we can use a HTML template instead of MD.

crutkas:
ref #720 as well.

crutkas:
@ralarcon issue i hit was I had to reference a final version to be able to see the text.  too much overlap and without reading a generated item, it is near impossible to understand what would be outputted.  I am also afraid at how this would localize with the smaller chunks of text.

ralarcon:
We can prioritize it as first after 1.2 issues discovered done if you agree. I think it is not so hard to convert it to a template with replacements

crutkas:
I defer to you guys, i'm on a bug filing spree right now :)  Before we submit files to loc, i think this should be fixed.

crutkas:
the gotcha here also is we are also mixing Markdown inside the resx along with the code behind.  We need to do #740 in the same motion.

sibille:
I think it's a good idea to have template for this. Will be easier to maintain and translate.

sibille:
I was looking at this yesterday and like the fact of using a resx template, but there is a drawback: resx templates do not support iterative or if logic. 
![image](https://user-images.githubusercontent.com/24698018/28307065-e43f5fae-6ba1-11e7-8d9d-5ba1a2187f2f.png)
For example: There is no way to not show the conflicting files section if there are no conflicting files. Also for the modified files section I had to create another template as I had to compose postaction code and strings 
![image](https://user-images.githubusercontent.com/24698018/28307155-35656ebe-6ba2-11e7-8c47-02a218fc5b07.png)
@crutkas: Could we use T4 templates for the summary generation, that would allow us to have one template per generated file?

mrlacey:
Because of the way the output is composed I'm not sure using a template defined in a resource is going to be efficient. We'd need multiple versions for each different combination of output.

Instead, I think it might be better to refactor the code to make it clearer what's happening. The current 92 line function could be broken up to be a lot more readable.

dgomezc:
Verified on dev-nightly
Wizard: 0.14.17249.65001
Templates: 0.14.17249.4

Issue #740, opened 2017-07-11
Duplication: CreateSyncStepsInstructionsPostAction and CreateSyncSummaryPostAction


crutkas:
Why do we have duplication.  Feels like CreateSyncStepsInstructionsPostAction and CreateSyncSummaryPostAction are the same.  I tracked this down based on looking at text

```
  <data name="SyncInstructionsTempFolder" xml:space="preserve">
    <value>You can find all referenced files in the temporary generation folder: {0}</value>
    <comment>s</comment>
  </data>

  <data name="SyncSummaryTempFolder" xml:space="preserve">
    <value>You can find the files from the generation in the temporary generation folder: {0}</value>
    <comment>s</comment>
  </data>
```

sibille:
The difference between the two is that CreateSyncStepsInstructionsPostAction is executed when you choose to integrate your files manually and CreateSyncSummaryPostAction is executed when the wizard merges changes from temp folder to project folder for you. 
The first text contains instructions for you to include the files in your project and all files, the second one is a summary of what has been done. 
I think it's important to have the two separated. Maybe I have to change the names a bit to reflect better what each does.

crutkas:
I didn't really see a big difference looking at the code.  I strongly believe the logic can be shared and maybe it is sending two different templates to the class.  Lets work together on rearchitecting this

sibille:
Yeah, logic is very similar, texts and files differ. Shared logic and different templates could work.

dgomezc:
Verified on dev-nightly
Wizard: 0.14.17249.65001
Templates: 0.14.17249.4

Issue #621, opened 2017-07-05
Right click summary file is created with non valid links.


crutkas:
Example from one.  Note the URL encoding.  When using the recommended editor, links are not clickable.

```
### Changes in file 'Strings\en-us\Resources.resw':
See the final result: [Strings\en-us\Resources.resw](c:%5Cusers%5Ccrutkas%5Cdocuments%5Cvisual%20studio%202017%5CProjects%5CApp22%5CApp22%5CStrings%5Cen-us%5CResources.resw)

```

sibille:
I think that's an issue of the editor. It's showing correctly all relative links but not the absolute ones. Maybe this was done on purpose, as the tool's purpose is to create html from markdown files (at least that's what I understood from the docs).   
If you open the file in Visual Studio Code you can navigate fine.

I see the following options to solve the problem: 

1. The generation summary file is currently created in the temp generation folder, if we create it in the project folder links could be relative and would work.
2. Just name the files and have no links. They are already part of the project so I think they won't be too hard to find. 
3. Find another extension that shows the files like Visual Studio Code.

crutkas:
I think we should figure out how to make it work with the recommended editor and not force someone to use VS Code (it is a great product) but asking someone to install another product just isn't realistic.

ralarcon:
Maybe we can look for other extension to show the md within VS. 

@crutkas What do you think about moving the file from the Temp folder to the project solution?

sibille:
Talking to the people from markdowneditor we found a way to generate a working link: 

```
### Changes in file 'Strings\en-us\Resources.resw':
See the final result: [Strings\en-us\Resources.resw](about:/C:/Users/sibil/documents/visual%20studio%202017/Projects/App178/App178/Strings/en-us/Resources.resw)
```
It's not a perfect solution:

*  the link is marked with a readline as unresolvable on the left side 
* the link does not work in other editors (f.ex. VS Code)

crutkas:
doing the shortened version which looks to be in the nightly works.

`[Strings\en-us\Resources.resw](Strings\en-us\Resources.resw)`

crutkas:
Actually, looks like #740 just bit us.  Pages works, features did not.

sibille:
The links in the CreateSyncSummaryPostAction were not clickable because they contained whitespaces. GenerationSummary.md is generated in the temp directory and the linked files are in your project folder which contains white spaces. 
[...]

PR #745, opened 2017-07-11
Combined logic, looked like one (feature versus page) worked but othe…

crutkas:
Fix for #621.  This really needs #740 and #741 in 1.3 as following the logic is hard to read and the duplication

### DIFF
--- a/code/src/Core/PostActions/Catalog/NewItemGeneration/CreateSyncSummaryPostAction.cs
+++ b/code/src/Core/PostActions/Catalog/NewItemGeneration/CreateSyncSummaryPostAction.cs
@@ -10,14 +10,11 @@
 // THE CODE OR THE USE OR OTHER DEALINGS IN THE CODE.
 // ******************************************************************
 
-using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Threading.Tasks;
+
 using Microsoft.Templates.Core.Gen;
-using Microsoft.Templates.Core.PostActions.Catalog.Merge;
 using Microsoft.Templates.Core.Resources;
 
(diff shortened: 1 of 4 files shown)


code/src/Core/PostActions/Catalog/NewItemGeneration/CreateSyncSummaryPostAction.cs
```diff
@@ -141,19 +134,14 @@ public override void Execute()
...
         {
-            return $"* {StringRes.SyncSummaryTempGenerationFile}: [{fileName}]({fileName}), {StringRes.SyncSummaryProjectFile}: [{fileName}]({FormatFilePath(filePath)})";
+            return $"* {StringRes.SyncSummaryTempGenerationFile}: [{fileName}]({fileName}), {StringRes.SyncSummaryProjectFile}: {GetMarkDownLinkToLocalFile(filePath)}";
         }
 
-        private static string FormatFilePath(string filePath)
+        private static string GetMarkDownLinkToLocalFile(string fileName)
         {
-            return $"about:/{filePath.Replace(" ", "%20").Replace(@"\", "/")}";
+            return $"* [{fileName}]({fileName})";
```
sibille: This will point to the file in the temp generation folder although it is already included in the project. 
Are we sure that we want to open the temp file and not the already integrated file? I think that could be weird and lead to confusion and errors (what if the developer does not notice and edits the temp file instead of the file integrated in the project)

ralarcon: I agree. If the changes are merged in the project, we should open the project file, not the external copy which would lead to confusion.  

@crutkas What do you think?

PR #763, opened 2017-07-12
Disable horizontal scroll on changes summary

mvegaca:


crutkas:
But why is this happening all of a sudden?  What shift there caused it to pop up?

PR #778, opened 2017-07-12
Fix licenses

mvegaca:


crutkas:
What prompted the change in code logic versus the need to just add in license.

mvegaca:
When I fix some licenses on templates I've seen that EmptyCollectionVisibilityConverter was not working correctly on wizard launched event and does not recognize dynamic changes on ObservableCollection properties. I implemented boolean properties to determinate if there are licenses, pages or features added and apply changes on the layout.

Issue #668, opened 2017-07-07
Project creation failed.


drchilds:
Possibly related to #609 

Stable release from the Marketplace, Template Studio projects work in VS2017 15.3 Preview, but not in the current 15.2 install. 

Solution folders are created, but are empty, and no Template Studio window appears.  Log follows:

```
>>>>>>>>>>>>>> Log started 20170707 05:06:24.371
>>>>>>>>>>>>>> Assembly File Version: 1.1.17173.2
[20170707 05:06:24.438]	dave	12884(1)		Information	Configuration file loaded: C:\PROGRAM FILES (X86)\MICROSOFT VISUAL STUDIO\2017\ENTERPRISE\COMMON7\IDE\EXTENSIONS\3ZSVAEFB.GSJ\WindowsTemplateStudio.config.json

>>>>>>>>>>>>>> Log started 20170707 05:19:47.862
>>>>>>>>>>>>>> Assembly File Version: 1.1.17173.2
[20170707 05:19:47.939]	dave	13948(1)		Information	Configuration file loaded: C:\PROGRAM FILES (X86)\MICROSOFT VISUAL STUDIO\PREVIEW\ENTERPRISE\COMMON7\IDE\EXTENSIONS\3ZSVAEFB.GSJ\WindowsTemplateStudio.config.json
[20170707 05:20:17.974]	dave	13948(1)		Information	Creating Windows Template Studio project...
[20170707 05:20:17.988]	dave	13948(1)		Information	Generating the template Default to c:\users\dave\Source\Repos\App8\App8.
[20170707 05:20:25.192]	dave	13948(1)		Information	Generating the template Blank to c:\users\dave\Source\Repos\App8\App8.
[20170707 05:20:27.627]	dave	13948(1)		Information	Generating the template Blank to c:\users\dave\Source\Repos\App8\App8.
[20170707 05:20:29.253]	dave	13948(1)		Information	Generating the template Settings to c:\users\dave\Source\Repos\App8\App8.
[20170707 05:20:31.226]	dave	13948(1)		Information	Generating the template Settings Storage to c:\users\dave\Source\Repos\App8\App8.
[20170707 05:20:32.796]	dave	13948(1)		Information	Generating the template Suspend and Resume to c:\users\dave\Source\Repos\App8\App8.
[20170707 05:20:33.914]	dave	13948(1)		Information	Generating the template Background Task to c:\users\dave\Source\Repos\App8\App8.
[...]

PR #781, opened 2017-07-12
Issue668 templating refs

ralarcon:
xref #668 

Added references to Microsoft.TemplateEngine 1.0.0.234 + Proper binding redirects.

Tested in vanila machine with 15.2 and 15.3 side by side


crutkas:
question on errors and what is displayed to customer and a code quality issue.

ralarcon:
@crutkas Nothing different than a crash will be shown to the user. This error should not be happening and if it happen then nothing will be operative. The UI is not initialized yet. The difference now is that if someone is affected, we are catching it in our telemetry.

Regading code quality issue, can you be more explicit?

Issue #873, opened 2017-07-21
FakeMsBuildProject.GetItemType, if VS switch statement


crutkas:
## Repro steps

i would expect this to be a switch statement:
```
if (ext == ".cs")
            {
                if (fileName.ToLower().Contains(".xaml.cs"))
                {
                    return VsItemType.CompiledWithDependant;
                }
                else
                {
                    return VsItemType.Compiled;
                }
            }
            else if (ext == ".xaml")
            {
                return VsItemType.XamlPage;
            }
            else if (ext == ".resw")
            {
                return VsItemType.Resource;
            }
            else
            {
                return VsItemType.Content;
            }
```

PR #884, opened 2017-07-25
adjusted to switch statement

crutkas:
## PR checklist

Quick summary of changes

#873 change to switch statement
that i did the flip correctly
yup

### DIFF
--- a/code/test/Fakes/MSBuild/FakeMsBuildProject.cs
+++ b/code/test/Fakes/MSBuild/FakeMsBuildProject.cs
@@ -112,30 +112,32 @@ namespace Microsoft.Templates.Fakes
 
         private VsItemType GetItemType(string fileName)
         {
-            string ext = Path.GetExtension(fileName).ToLower();
-            if (ext == ".cs")
+            VsItemType returnType = VsItemType.Content;
+
+            switch (Path.GetExtension(fileName).ToLower())
             {
-                if (fileName.ToLower().Contains(".xaml.cs"))
-                {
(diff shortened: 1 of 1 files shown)


code/test/Fakes/MSBuild/FakeMsBuildProject.cs
```diff
@@ -112,30 +112,32 @@ private bool ItemExists(string itemPath)
...
-                return VsItemType.Resource;
-            }
-            else
+            VsItemType returnType = VsItemType.Content;
+
+            switch (Path.GetExtension(fileName).ToLower())
             {
-                return VsItemType.Content;
+                case ".cs":
+                    if (fileName.ToLower().EndsWith(".xaml.cs"))
```
dotMorten: I'd suggest using `fileName.EndsWith(".xaml.cs", false, CultureInfo.Invariant)` instead of using ToLower.
https://msdn.microsoft.com/en-us/library/t9h2fbth(v=vs.110).aspx

crutkas: Should do line 117 with same fix as well

dotMorten: you can't really do invariant comparison in the switch cases, unless you use the latest C# features, but in that case I think tolower up front for reuse in all the cases is actually better

crutkas: with tolower, you can apply invariant culture

crutkas:
@dotMorten updated

Issue #891, opened 2017-07-26
using System; should be in every file, is now missing in the default generated code behind file


mahoekst:
## For Bugs:
Your code should be set up to work for developers to get started. Using System; is part of that although your pages and code might not need it.
### Repro steps
For example do this: 

File new, next next next finish (so just 1 basic page)
Add button on mainpage
Double click on button
Paste this code inside the click handler. Make the handler async

UserConsentVerificationResult consentResult = 
              await UserConsentVerifier.RequestVerificationAsync("userMessage"); 
if (consentResult.Equals(UserConsentVerificationResult.Verified)) 
{
// continue 
}

Add: using Windows.Security.Credentials.UI;

Compile

Severity | Code | Description | Project | File | Line | Suppression State
-- | -- | -- | -- | -- | -- | --
Error | CS4036 | 'IAsyncOperation<UserConsentVerificationResult>'   does not contain a definition for 'GetAwaiter' and no extension method   'GetAwaiter' accepting a first argument of type 'IAsyncOperation<UserConsentVerificationResult>'   could be found (are you missing a using directive for '**System**'?) | UserConsentVerifier | C:\Users\mahoekst\Downloads\Windows-universal-samples-master\Windows-universal-samples-master\Samples\UserConsentVerifier\cs\Scenario2_RequestConsent.xaml.cs | 45 | Active

Although the error states about system, I missed it the first time and wasted time. My guess most (beginning) developers would miss this too.

### Expected Behavior

My app should just compile

### Actual Behavior

Severity | Code | Description | Project | File | Line | Suppression State
-- | -- | -- | -- | -- | -- | --
Error | CS4036 | 'IAsyncOperation<UserConsentVerificationResult>'   does not contain a definition for 'GetAwaiter' and no extension method   'GetAwaiter' accepting a first argument of type 'IAsyncOperation<UserConsentVerificationResult>'   could be found (are you missing a using directive for '**System**'?) | UserConsentVerifier | C:\Users\mahoekst\Downloads\Windows-universal-samples-master\Windows-universal-samples-master\Samples\UserConsentVerifier\cs\Scenario2_RequestConsent.xaml.cs | 45 | Active

mrlacey:
This is now in the dev-nightly build.

PR #914, opened 2017-08-07
Ensured `using System;` now in every generated file

mrlacey:
#891


crutkas:
would have assumed more would have been there.

Issue #885, opened 2017-07-25
Navigation Template Back Button Doesn't Restore Storyboard State


michael-hawker:
### Repro steps

1) Create a Project with a Navigation Pane and a Couple of Pages.
2) Run the project and click on a couple of pages.
3) Now click the back button in the top-left.

Note: Previous page still has a highlight bar to it's left:

![image](https://user-images.githubusercontent.com/24302614/28596484-98dbcbcc-714d-11e7-8dbc-7431c5ec7cb4.png)

App does seem to switch page and think the current one is selected, but there's some lingering visual state remaining.

### Expected Behavior

Expect Previously Selected Highlight bar on left to disappear.

### Actual Behavior

What really happened

## System

* **VS Version:** 2017 15.2 26430.16
* **WTS Wizard Version:** 1.2.17198.1
* **WTS Template Version:** Not sure where to get this?
* **Windows Build:** 15063

mrlacey:
This feels very familiar and like something I addressed before when changing the styling of the navigation menu items.
Possibly a regression when switching to the updated community toolkit control.
investigating....

mrlacey:
Yes, it all comes back to me now.
It's a bug in the HamburgerMenu control. Although [they say it's not a bug](https://github.com/Microsoft/UWPCommunityToolkit/issues/969#issuecomment-281396945).
The control includes an indicator of the selected item (internally it's even called `SelectionIndicator`) but it doesn't handle the back button. (I guess they don't count this as "selection".)
The reason that two items appeared to be selected is that one was being correctly indicated by us and the other was incorrectly being indicated by the control. Normally our control appears over the top of theirs but when they're out of sync you see two.
I think it's their bug as they are showing an indicator which we can't control and works most of the time but when it doesn't they say we're "responsible for selection synchronization" but we don't use the underlying selection system as it's not appropriate and doesn't work when we have two lists (like we do).
Anyway...

**workaround**

Add a background to the Grid in the DataTemplate at the top of ShellPage.xaml. Like this:

```xml
<Page.Resources>
  <DataTemplate x:Key="NavigationMenuItemDataTemplate" x:DataType="vm:ShellNavigationItem">
    <Grid Width="320" Height="40" Background="{ThemeResource SystemControlBackgroundAltHighBrush}">
```

This hides the indicator that they show and so the only one that's ever seen is ours, which we know is always correct because it's synchronized with what's displayed on the main part of the view and also reflects the text color highlighting.

[...]

PR #931, opened 2017-08-11
Ensure native Hamburger selection highlight is hidden

mrlacey:
#885

deliberately against master for hotfix


crutkas:
ok, confirmed your fix.  With #350 now in, we need to validate fixes here for dev but this does fix master.

crutkas:
going to wait for CI to pass then merge in.

crutkas:
Forcing a build, opened bug for not having one auto kick off

PR #938, opened 2017-08-11
Fix for pivot life cycle.

crutkas:
## PR checklist

Quick summary of changes

#350
Pivot fix

can't run full tests since something in dev branch broke core.test


crutkas:
@mrlacey can you review so this fix can get in?

Issue #877, opened 2017-07-23
Should Order.cs from SampleDataService be more clear it is a sample data


crutkas:
**Overview:** in reference to #870, should Order.cs be actually "SampleOrder.cs" or some way to more clearly outline it is a sample/mock class.

mrlacey:
The linked issue was about some text that was actually wrong. This is more subjective. As there's already a comment in there that says it's meant to be deleted as it only exists for the sample data I'd think that's enough but as you're asking the question it demonstrates that not everyone thinks the same way on this. In that case, yes, change it. Making the change will be quicker than any further discussion and I can't see a downside to doing so.

mrlacey:
verified in dev-nightly
wizardVersion="0.14.17228.65001"
templatesVersion="0.14.17228.1"

PR #941, opened 2017-08-12
Rename Order[.cs] to SampleOrder[.cs]

mrlacey:
#877


mrlacey:
observer type in branch name. this is for #877, not 887

crutkas:
this is actually for #877

Issue #849, opened 2017-07-20
Why is Templatex called this?


crutkas:
Why is this file called Templatex?  Should it be TemplateEx?  Or it looks to be doing Template Packing, unpacking and Signing, maybe call the file something in lines with that?

mrlacey:
The class could also be made static as it only contains constants and static methods.

crutkas:
i just did a bunch of clean up on this where we had dup code versus calling into one another. 

I feel like this should be in Core.IO.Packaging and called MstxFile

crutkas:
@ralarcon and @sibille any ideas?

ralarcon:
No special reason behind the scenes... I personally set the class name to Templatex as it was dealing with the packaging of templates (the suffix "x" is used in multiple places where packs things using _Open Packaging Conventions_ -docx, appx, etc.). You can rename it and move to the mentioned namespace...

crutkas:
this is ready for someone to review, FYI

PR #952, opened 2017-08-13
Templatex rename and making into static class

crutkas:
#849 improvements minus renaming / locating the file.  less duplicated code, making sure methods call their parent method versus duplicate code.

Monday i'll run unit tests against this as looks like xunit pooped the bed on my laptop so currently i'm unable to verify.


crutkas:
unit tests run like a champ

crutkas:
Testing fixes for this on a different box that still have 15.2 installed.  not safe to merge now that i adjusted naming

crutkas:
Tests pass, we can merge when ready

Issue #692, opened 2017-07-10
centralize datetime tostring writing


crutkas:
#674 improvement

crutkas:
Found 
- [ ] ShellHealthWriter
- [ ] FormatWriterMessage

mrlacey:
Also in 
- FormattedWriterMessages
- ShellHealthWriter
- GenContext
- GenerationFixture
- NewItemGenerationTests
- StyleCopGenerationTestsFixture
- VsOutputPane

mrlacey:
ReOpening as `FormatAsFullDateTime` would be improved with a hyphen separating the year, month, and day parts of the date so it's easier to read.

mrlacey:
Before
![image](https://user-images.githubusercontent.com/189547/29669648-0a52df0c-88dc-11e7-81db-569bd2f8e638.png)

After
![image](https://user-images.githubusercontent.com/189547/29669809-8d440d8c-88dc-11e7-8e63-e255ab7a4d8f.png)

PR #959, opened 2017-08-15
centralize datetime tostring writing

mrlacey:
For #692
Also ensures that Hours are in 24 hour format (for #674)


mrlacey:
Before
image

After
image

ralarcon:
Good!

PR #970, opened 2017-08-18
Include explicit Fluent Design features in Webview and M/D templates

lucastaiMSFT:
This PR serves to introduce fluent features in the WTS default templates for inclusion in the FCU update. 

1) It updates the Webview template with an acrylic control bar and the Master/Detail template with connected animation and an implicit show animation.
2) It adds new fields to the Order model in the SampleDataService so each order has a unique icon.
3) It serves to update the Master/Detail template closer to a Design reviewed style. 
4) It adds a documentation page detailing how to add Perspective Parallax to a template. 

Because this is utilizing features (Conditional XAML) that only become available in RS2 and later, compiled projects from these templates will not work in any version of Windows below RS2. There is also additional design review internally that should be done to verify that the M/D template is correct and fits redlines for our purposes and that the use of Acrylic in Webview is warranted.

### DIFF
--- a/docs/Fluent_PerspectiveParallax.md
+++ b/docs/Fluent_PerspectiveParallax.md
@@ -1,0 +1,99 @@
+# Perspective Parllax
+
+## Overview and Purpose
+
+This page describes how to enable Fluent Design parallax features in a
+Windows Template Studio enabled application.
+
+## Prerequisites
+
+Perspective Parallax as a control is available as of the Fall Creators Update,
+which enables all of the new Fluent features. This method also utilizes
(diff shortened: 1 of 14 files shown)


mrlacey:
Does this merge mean that when we incorporate this with the main branch we're going to only support CU & FCU?

crutkas:
Yup. Starting When FCU is released, we shift the supported versions to FCU and CU. currently we support CU and AU.  The current release and the release prior. 

This has been our public pledge from the start.

mrlacey:
In my head, I was thinking we supported N-2. My mistake. Thanks for clarifying.

Issue #727, opened 2017-07-11
PNG and "highlighting" in wizard


crutkas:
The MVVM Light logo is actually a PNG.  It would be great if we could get that to highlight to the correct color.  Could a WPF shader effect do this?

crutkas:
here is some stuff that could be an item but man, rathole to get set up.

http://wpffx.codeplex.com/
http://wpf.codeplex.com/releases/view/14962
http://wpffx.codeplex.com/discussions/37107

mrlacey:
Or we could just have two versions of the image and swap between then as necessary

crutkas:
high contrast mode is where this could cause a curve ball or if we change this to accent color

crutkas:
synced with @onovotny and he recommended looking at https://github.com/teichgraf/WriteableBitmapEx/

This could do it and i'm pretty sure it is a nuget package too.

crutkas:
using writable bitmap, have a functional version.

```
private void Image_MouseEnter(object sender, MouseEventArgs e)
{
    var setColor = Colors.CornflowerBlue; // highlight color
    ChangeColorOfImage(sender, setColor);
}

private void Image_MouseLeave(object sender, MouseEventArgs e)
{
    var setColor = Colors.Black; // base brush color
    ChangeColorOfImage(sender, setColor);
}

private static void ChangeColorOfImage(object sender, Color setColor)
{
    var img = sender as Image;
    WriteableBitmap writeableBmp = new WriteableBitmap(img.Source as BitmapSource);
    img.Source = writeableBmp;

    using (writeableBmp.GetBitmapContext())
    {
        writeableBmp.ForEach((x, y, color) => Color.FromArgb(color.A, setColor.R, setColor.G, setColor.B));
    }
}
```

PR #975, opened 2017-08-18
enable bitmap effects

crutkas:
## PR checklist

Quick summary of changes

#727 
perf

### DIFF
--- a/code/src/UI/Controls/ImageEx.cs
+++ b/code/src/UI/Controls/ImageEx.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Templates.UI.Controls
     public class ImageEx : ContentControl
     {
         private const string XamlExtension = ".xaml";
+        private Color _updatedColor;
 
         public override void OnApplyTemplate()
         {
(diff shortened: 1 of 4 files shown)


code/src/UI/Controls/ImageEx.cs
```diff
@@ -50,7 +73,7 @@ private UIElement CreateFromBitmap()
             var image = new Image
             {
                 Source = bitmap,
-                Stretch = Stretch
+                Stretch = Stretch,
```
mrlacey: Having commas on the end of the last line is recommended as it means that adding a line after it means you only have to add the new line, not also change the last old line too.

code/src/UI/Controls/ImageEx.cs
```diff
@@ -70,7 +93,7 @@ private UIElement CreateFromXaml()
 
                 if (paths.Any())
                 {
-                    paths.ForEach(p => BindingOperations.SetBinding(p, Shapes.Path.FillProperty, CreateBinding(this, nameof(Foreground))));
+                    paths.ForEach(p => BindingOperations.SetBinding(p, Shapes.Shape.FillProperty, CreateBinding(this, nameof(Foreground))));
```
mrlacey: I'm curious why this change? It doesn't change any behavior but just move a reference up the inheritance hierarchy. On its own, this line still looks fine, but in context, it's next to a block of code that works on "Shapes" and so the block of code about "Paths" now has a mention of "shapes" that could be potentially confusing and looks like a cut and paste error.

code/src/UI/Controls/ImageEx.cs
```diff
@@ -36,12 +37,34 @@ public override void OnApplyTemplate()
...
+                writeableBmp.ForEach((x, y, color) => Color.FromArgb(color.A, setColor.R, setColor.G, setColor.B));
             }
         }
 
         private UIElement CreateFromBitmap()
         {
-            var bitmap = CreateIcon(SourcePath);
+            var bitmap = new WriteableBitmap(CreateIcon(SourcePath) as BitmapSource);
+
             if (bitmap == null)
```
mrlacey: This check will always be false and so isn't needed.

Issue #988, opened 2017-08-22
StyleCop rule CA2214 violations


ralarcon:
The `StyleCopPlusLocalTemplatesSource` and `ProjectTemplateGeneratorBase`clases are violating the rule [CA2214](https://msdn.microsoft.com/library/ms182331.aspx)

@mrlacey I would like to dobule check with you so I'm ignoring the warings by now until further discusion to check if a code refactor is needed.

ralarcon:
The class `BaseFileViewModel` also violates this rule.

crutkas:
I'm fine pushing this back to 1.4 FYI

mrlacey:
To save me having to look this up again next time I view this

[CA2214: Do not call overridable methods in constructors](https://msdn.microsoft.com/en-us/library/ms182331.aspx)

mvegaca:
Fixed on #1230

PR #989, opened 2017-08-22
Ignore CA2214 until further review

ralarcon:
## PR checklist

 #988
No
Yes


crutkas:
@ralarcon should this have been added to the global suppression list instead?

ralarcon:
I did it in the source code with the intention to have a revision so, if finally we are suppresing it, we will move it to the suppresion file, if not we need to make some chanages. In other words, it was done intentionally.

Issue #957, opened 2017-08-15
Reinstate all unit tests for project generation


crutkas:
For #946, we disabled large sums of tests.

We disabled

- GenerateAllPagesAndFeatures
- GenerateProjectWithIsolatedItems
- GenerateEmptyProject

ralarcon:
Moving this 1.3 as part of the CI / Deploy builds enhancenments.
Reviewing the approach for integration tests for generation:
* I will split the testing using Traits / Collections to allow to execute just Generation tests (quick run) and build the generated content (slow due to the build process it self).
* Generation tests will be part of the CI which will ensure that, at least, the generation works fine
* Building the generated projects (slow and time consuming) will be executed every time we commit code (lets try to do it just when the templates are modified -not sure if possible). This build test will run in a dedicated build server to avoid blocking developers.

mrlacey:
Can changes made here be done in connection with #831 so that it becomes clear on the impact of adding new tests in different places? And even the consequences of adding new test projects.

ralarcon:
No are working in CI / Integration builds

PR #998, opened 2017-08-24
Re-Include OneByOne integration tests

ralarcon:
We have setup a custom build server for these test which are really time consuming. 
The nature of these tests (build real apps) make it difficult to reduce the time required as they are combining all possible page and features templates with project and framework types. Next work arround this will be to try to improve the performance by parallelizing.


crutkas:
this is a fix for #957

Issue #506, opened 2017-06-23
Feature: scheme launch to app


crutkas:
This will allow an application to to define a custom scheme URI and allow for a custom launch.  It should have stubbed area for parsing the query string in the area where it is handled.

Example would be `Contoso:?Addresses={address}&Body={body}`

https://docs.microsoft.com/en-us/windows/uwp/launch-resume/handle-uri-activation
https://docs.microsoft.com/en-us/windows/uwp/launch-resume/launch-default-app 

This would be an extension of #180.

mrlacey:
Isn't this how a Cortana Skill (#145) would do this?

mrlacey:
Also, this will need the ability to capture some extra information in the wizard. Just like as discussed for an enhancement (#325) or adding file headers (#477)

crutkas:
Actually, yup, #145 would be how we'd pull that off

mrlacey:
Having thought about how to implement this, I'm looking for feedback on plans (if seem right, not how to do this) to implement the following:

- The value for the protocol will be asked for in the wizard in the same way the name of a page can be specified. (New functionality for the validation rules and failure error message for this will be added to the template config)
- This feature will add another page (called ProtocolExample) to show passing information from a URI to a page within the app. This page is purely for demonstration and will include comments saying that it is for sample purposes only and should be removed
- For NavPanel and Tabs&Pivot projects, the activation handler will navigate to the example page within the shell/tab pages and have comments for how to navigate to a page, not within the shell
- The example page will display parameters passed to it (if any)
- Comments in the top of the activation handler will list supported protocol content formats
- The activation handler will show a basic example of querying the arguments for valid content in the CanHandleInternal method
- The current activation document will be extended with more information about protocol launching

touseefbsb:
@mrlacey  the way you described seems reasonable and perfect :)

mrlacey:
This is going to require a change to the wizard. The change will mean the new template won't work with the old wizard. Do we have a way of ensuring that a template can only be used with a certain version (or above) of the wizard? Or is there another way to stop anyone who doesn't update the VSIX have issues once the new template is public?
 //cc @ralarcon  @crutkas

mrlacey:
bumping to 1.3 due to lack of time to integrate wizard change before 1.2 is due

crutkas:
agreed

ralarcon:
[...]

PR #999, opened 2017-08-24
New feature: URI Scheme (protocol) support

mrlacey:
#506


crutkas:
i looked at this, looks good, i want to run it before i approve

crutkas:
@mrlacey ran it, looks good.  i thought the idea of a hidden page was interesting.  would a message prompt make this feature a lot lighter versus an entire new page?  or is it you want to demonstrate the e.Parameter usage?

I'm also going to have app model team give it a quick once over as well.

mrlacey:
@crutkas My expectation is that a major use case is using a protocol launch to navigate to specific pages within an app. (It's core to Rome Web-to-app and in many Cortana skill examples.) Beyond the home page, there's no page guaranteed to be in the created app and so it requires the inclusion of an extra page to show this functionality. 
Using an extra/hidden page was always my [plan](https://github.com/Microsoft/WindowsTemplateStudio/issues/506#issuecomment-310859649), for the purpose of demonstrating how to pass parameters from the Uri Query and for completeness. In that there will be people who want to do this it's easier to have the people who don't ignore or remove it than to add more documentation to try and explain how.

crutkas:
Got the review, approach looks good.  Per before, i think navigating to an extra page adds a bit of load on us but I understand the reasoning.

PR #1003, opened 2017-08-24
Create InformationButton control

mvegaca:


crutkas:
@mvegaca Needs header on CS file before this can be merged in.  How did the analyzer not catch that?

mvegaca:
You're right @crutkas I've fix the analyzer and commited

mrlacey:
The change doesn't "fix [the] analyzer" it fixes what the analyzer should have reported as a warning.
@mvegaca Was this not reported on your machine? If the StyleCop Analyzer isn't running for you it may not also run for others and end up creating lots more work.

PR #1014, opened 2017-08-28
Fix async methods on vs emulator

### LINKED ISSUE #1013: VS Emulator does not work
VS Emulator can not create a CSharp project

mvegaca:
Fix #1013

### DIFF
--- a/code/test/VsEmulator/Main/MainViewModel.cs
+++ b/code/test/VsEmulator/Main/MainViewModel.cs
@@ -146,14 +146,14 @@ namespace Microsoft.Templates.VsEmulator.Main
             SolutionName = null;
         }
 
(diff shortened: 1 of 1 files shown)


crutkas:
These are async functions that need to have their names updated.  `private async void NewCSharpProjectAsync()`

Shouldn't our newly added analyzers caught this?

mrlacey:
I think I recall leaving out VSEmulator from the projects this was included in as the analyzer is strict about a lot more than just this and applies to some very specific functionality within VS that doesn't really apply to the emulator.
There's no reason it couldn't be included though.

> Install-Package Microsoft.VisualStudio.Threading.Analyzers -Version 15.0.240

Issue #1012, opened 2017-08-28
NavigationService updates


mvegaca:
Working on #775 issue, I built a [MixedNavigationSample](https://github.com/mvegaca/WindowsTemplateStudio/tree/Issue775-documentation/samples/navigation/MixedNavigationSample) that is an app that starts on a Login page with username and password textboxes and a login button. When press login button the applications navigates to a Splitview page with home and settings navigation items.

![navigation](https://user-images.githubusercontent.com/5556784/29769442-b4723d14-8bea-11e7-800f-44efd5a49616.png)

When I navigated from LoginPage to ShellPage I show that Application lost goback functionality.
This is because ActivationServices is registered to NavigationService.Frame navigation methods on app initialization and after doing login, ShellPage replaces current frame for a shellpage frame.
To solve this, I've added in sample own navigation event handlers to navigation service. In this case, ActivationService is subscribed to those event handlers and this subscription does not break when Frame is replaced.

I'm going to try this on templates using different frameworks.

mvegaca:
Added NavigationService updates on #1016

Here on this point, I would like to work on a PoC to build a page template (i.e. LoginPage) that can navigate or be navigated to a shell page, and think about the postactions, implication and breaking changes that this template could have.

mvegaca:
@crutkas how about I work LoginPage template between app launch and home page?

mrlacey:
There's already a feature request for creating a login page #291

mrlacey:
It seems there's some overlap here with #697. Does this make that irrelevant?

crutkas:
we should not have a sample with a undefined feature like login page.  Please do not do that in a sample.  More so samples provide guidance.  If our sample provides a login page, it should be correct.  I would actually provide a more abstract example versus what you proposed. 

We really need to discuss samples and guidance for stuff like this so we are inline with true guidance someone may reproduce.

mvegaca:
Code separated on two different pull request:
- Navigation updates for templates #1016
- Navigation [documentation ](https://github.com/mvegaca/WindowsTemplateStudio/blob/navigation-sample/docs/navigation.md)and sample #1036

I've updated the Mixed navigation sampled to change LoginPage for StartPage.

![Mixed Navigation Sample](https://user-images.githubusercontent.com/5556784/29875725-2349b8f0-8d9b-11e7-9dc5-ecf37ff1c892.png)

mvegaca:
Add missing event references on composition templates #1086

[...]

Issue #775, opened 2017-07-12
Document how to navigate between blank page and navigation pane.


Tostifrosti2:
When you create a new project with Windows Template Studios, you'll get a wizard. In this wizard, you can choose between 'Navigation Pane' and 'Blank Page'. Why isn't it possible to merge those options together?

In real world scenario, you may want to start with a login page and after logging in, you will be navigated to the 'Navigate Pane' page.

mrlacey:
@Tostifrosti2 We already have plans for introducing the ability to add a login page as the first page seen in issue #291 

The **Blank** project type is intended for if you want to completely control the structure and in-app navigation yourself. Your request for a

Until this is built in I expect it'll be beneficial to add some docs to explain how to do this yourself.  
I'm going to change this issue to be about adding the documentation and it should get done soon.

Tostifrosti2:
@mrlacey Thank you for the quick response!

I have already started a project where I used the MVVM Light as start-up project.
I noticed that the NavigationService is used to switch between pages using the Window.Current.Content as Frame (that was initialised in the ActivationService). 

I've tried to create a method myself to set the frame, but with no success.

`public bool NavigateToFrame(string pageKey, UIElement frame, object parameter = null, NavigationTransitionInfo infoOverride = null)
        {
            lock (_pages)
            {
                if (!_pages.ContainsKey(pageKey))
                {
                    throw new ArgumentException($"Page not found: {pageKey}. Did you forget to call NavigationService.Configure?", "pagekey");
                }
                if (frame != null)
                {
                    Frame.BackStack.Clear();
                    Window.Current.Content = frame;
                    _frame = null;
                }
                var navigationResult = Frame.Navigate(_pages[pageKey], parameter, infoOverride);
                return navigationResult;
            }
        }`

This method above will result in a NullpointerException, because the Frame is null. The weird thing is, is if I debug this code above, the application just runs fine. :S
Could you help me out?

Tostifrosti2:
@mrlacey Thank you for your help, but i've already solved the problem.

I've changed the method a bit and overridden the 'OnNavigatedTo' method in the MainPage (Navigation Page).
In this 'OnNavigatedTo' i give the Frame to the ViewModel so the ViewModel can initialize the Frame in the NavigationService. 

NavigationService.cs
```
public bool NavigateToFrame(string pageKey, UIElement frame, object parameter = null, NavigationTransitionInfo infoOverride = null)
        {
            lock (_pages)
[...]

PR #1016, opened 2017-08-28
NavigationService updates

mvegaca:
Related to #1012 
Add navigation events in navigation service to templates.
Reviewed and passed al tests


crutkas:
Can we break the sample out from the template code update PR?  it is hard to break the two apart and i think it is critical to review the template updates.

It is also unclear why both #1012 and #775 both refer to documentation updates yet this has actual template updates.

templates/_composition/CodeBehind/Project/Services/NavigationService.cs
```diff
@@ -51,5 +60,24 @@ public static bool Navigate(Type pageType, object parameter = null, NavigationTr
         public static bool Navigate<T>(object parameter = null, NavigationTransitionInfo infoOverride = null)
             where T : Page
             => Navigate(typeof(T), parameter, infoOverride);
+        
+        private static void RegisterFrameEvents()
+        {
+            _frame.Navigated += _frame_Navigated;
+            _frame.NavigationFailed += _frame_NavigationFailed;
+            _frameEventsRegistrated = true;
```
crutkas: if we set this here, why isn't it unset in unregister?

templates/_composition/MVVMBasic/Project/Services/NavigationService.cs
```diff
@@ -51,5 +59,24 @@ public static bool Navigate(Type pageType, object parameter = null, NavigationTr
         public static bool Navigate<T>(object parameter = null, NavigationTransitionInfo infoOverride = null)
             where T : Page
             => Navigate(typeof(T), parameter, infoOverride);
+        
+        private static void RegisterFrameEvents()
+        {
+            _frame.Navigated += _frame_Navigated;
+            _frame.NavigationFailed += _frame_NavigationFailed;
+            _frameEventsRegistrated = true;
```
crutkas: if we set this here, why isn't this unset in unregister

templates/_composition/MVVMLight/Project/Services/NavigationServiceEx.cs
```diff
@@ -85,5 +93,24 @@ public string GetNameOfRegisteredPage(Type page)
                 }
             }
         }
+
+        private void RegisterFrameEvents()
+        {
+            _frame.Navigated += _frame_Navigated;
+            _frame.NavigationFailed += _frame_NavigationFailed;
+            _frameEventsRegistrated = true;
```
crutkas: if we set this here, why isn't this unset in unregister

crutkas:
Thanks for breaking this out, why do we set but never unset the bool for reg'ing the events?

Why do we even care, best of my knowledge, you can remove events that were never wired up with no issue.

mvegaca:
Your are right @crutkas thank you very much.

crutkas:
Why do we still have that bool check.  It *could* double up registering the events but you check to unreg?  wouldn't you check if the events are already wired as well?

mvegaca:
You are right @crutkas it was enough with frame not null validation

Issue #757, opened 2017-07-12
Info prompt bar disappears too quickly at times


crutkas:
I can't repro this consistently but i've seen it multiple times.

**Repro steps that i've seen it occur:**
1. select framework and project type
1. click next
1. add page
1. click back
1. change framework

**expected:**
bar will appear and go away in ~5 seconds

**actual:**
bar flashes on the screen for a ~half a second

crutkas:
How i can get it to repro again and again:

follow repro steps (you don't actually need to add a page)

To cause it to occur again, fully exit WTS and restart your file->new experience

mvegaca:
Now I can reproduce it, I'm going to work on it.

mvegaca:
This occurs because when the template update process ends wizard status is cleaned and in this case, current status is an important warning.
Added a PR #1017 with current status can be cleaned validation.
Can you approve the PR @crutkas?

crutkas:
punt to 1.4

mvegaca:
Fixed on #1236

mvegaca:
Reopening

crutkas:
why did the fix seem like it worked but didn't?

ralarcon:
The fix is inplace but we didn't verified it yet with the nightly version... previously closed by error.

sibille:
Verified on dev-nightly:
Wizard: 0.14.17283.65001
Templates: 0.14.17283.1

PR #1017, opened 2017-08-28
Fix status cleaning

mvegaca:
Check if current Status in wizar can be cleaned before clean it.

Quick summary of changes
#757


crutkas:
the naming is unclear, totally not sure what this does

mvegaca:
The problem was that in the wizard there are processes that can update or clean the top status bar, in this case, templates update process was cleaning the status when it finishes with no pending updates. I've added a property in status that specifies if this status could be cleaned (information status are cleanable and warning and error status not). Can you approve the PR to can confirm the solution on dev nightly?

mrlacey:
Is this code using the word "clean" instead of "clear"?

mvegaca:
CleanStatus is when the status is setted as empty, for example, "looking for templates..." -> " "

mrlacey:
Yes, but that's not what "clean" means

crutkas:
I'm trying to wrap my head around this, this feels awkward from an implementation standpoint.  It is unclear how the fix here actually corrects #757 as 757 feels like a race condition where something is causing it to close early due to an older timer or condition or something being triggered.

As a dev, if i called `StatusViewModel.Error("foo", true, 15);` versus `StatusViewModel.Error("foo", false, 15);`.  Or `StatusViewModel.Error("foo", false, 0);` and then called `StatusViewModel.Error("bar", false, 0);` i don't know what exactly would happen.

When i think about a dialog as well, it is either Transient or Persistent dialog

- A transient message disappears based on time (or something else)
- A persistent message would stay up until a direct action corrects it (error fixed, user closed, ...)

When would either of these states not be cleared out when it goes away or another warning super-succeeds it.  I think the type of dialog also would be thought about for our we handle our markdown modals for.  (that is a different topic however).

Naming wise as well, i agree with Matt, you're clearing, not cleaning.  Cleaning would be something like destroying references and closing streams.

mvegaca:
Naming updates added. For another release, I would like to think about a different and unified message display method, that allows having a status pool to read all notifications, with some improvements. If there are one status showing and one thread wants to show a new status, this new status keep in a pool until the current message is autohide or closed, defining different priorities.
Can @crutkas approve the PR and merge the code?

crutkas:
@mvegaca, there is a conflict, can't merge

Issue #967, opened 2017-08-18
Click Cancel Button on the Wizard blows up VS2017 15.3


EdiWang:
### Repro steps

1. Open An Exisiting Solution.
2. Right Click on the Solution file and Add a new Template Studio project.
3. When the first screen of the wizard loaded, click "Cancel" button.

### Expected Behavior

Nothing would happen to the current IDE screen.

### Actual Behavior

An error dialog show up, says:

"The process cannot access the file 'storage.ide' beacuse it is being used by another process."

## System

* **VS Version: 2017, 15.3**
* **WTS Wizard Version:1.2.17198.1**
* **WTS Template Version:1.2.17202.1**
* **Windows Build:10.0.15063.540**

crutkas:
I can't reproduce this.  Not saying it isn't happening.

Does it happen every time you do this?

EdiWang:
@crutkas , I am able to reproduce it every time on both of my machines. I recored a viedo for that, please check "https://www.youtube.com/watch?v=TDyb3foOq7g". 

And it is happening to VS2017 15.3.1 also, I got this VS update just this morning.

crutkas:
@EdiWang that video was SUPER helpful.

@ralarcon any ideas?

ralarcon:
Taking a closer look...

mrlacey:
I suspect this is related to a VS extensibility reference issue as using the 1.2 wizard and VS 15.3

crutkas:
Nightly looks to have this fixed.

Issue #882, opened 2017-07-25
Cancelling a "Add new project" into exist project with WTS deletes directory


ray1997:
Today, I'm just try to add new project but I did not notice it select Windows Template Studio
so I name it and click ok, But when wizard window show up I tried to cancel it, But I get use to how most app have cancel on bottom right corner. So I press it *twice* without looking that I'm press next and it creating new project. So I wait.. But it quite too long so I tried to check project folder for anything else. But then I saw my project folder has only "obj" and "bin" folder left.
And I tried creating new project again. And it still delete test project. :(

crutkas:
For the WTS deleting everything, I can't reproduce this however.  Can you give step by step details on reproducing?  Remember a solution and a project are different items too.  You can't add a project to a project.

Trying to repro this, I did:

1. file->new and did a console app.
2. Right click on the solution
3. Add->New Project
4. select Windows Template Studio
5. click through wizard quickly
6. solution has console project and WTS project.

I used

- WTS Version: 1.2.17198.1
- Template Version: 1.2.17202.1

For our cancel button placement, that is 100% fair and valid feedback looking at how Visual Studio puts their OK versus cancel button.  I think that is a different issue and should be a logged as such.

ray1997:
I'm afraid I can't reproduce it anymore.
I swear this is not a joke. I did lose my project files. On both real project and test project.

But maybe there is some other issue that not from WTS. All I can remember is I just install WTS along with Template 10 and Multilingual Toolkit and after I relaunch app. I just add new project (click on solution) And the just happen.

I'm sorry, Maybe it's something else. I'll close this issue now.

![sketch 1](https://user-images.githubusercontent.com/14814401/28588429-bdd879b4-71a4-11e7-8af2-2a6675be80b3.png)

crutkas:
Ray, I'm not doubting you, i need more information so we can see.  Is the steps i did above what you did?  What was the other project(s) in your solution?

I can say this was a scenario I never tested as I always assumed file->new, not adding to an existing solution new project

ray1997:
In my case both project is UWP project. And I added more UWP project by right click on my solution and add new project with WTS project
Some info about my project: There are two exist project in solution and both of them is also UWP project.
but that maybe not relate as test project has only 1 project but it still delete everything
More info: I was on Windows insider fast ring build 16241 and install visual studio community 2017 with version 15.2 (26430.16)
[...]

PR #1019, opened 2017-08-28
Delete solution directory to cancel button click, only if directory is empy.

dgomezc:
## PR checklist

Quick summary of changes

#882  #967

### DIFF
--- a/code/src/UI/VisualStudio/SolutionWizard.cs
+++ b/code/src/UI/VisualStudio/SolutionWizard.cs
@@ -121,6 +121,5 @@ namespace Microsoft.Templates.UI.VisualStudio
                 Directory.Delete(solutionDirectory, true);
             }
         }
-
     }
 }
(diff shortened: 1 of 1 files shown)


crutkas:
how does this fix the storage.ide issue?  Should we increase any unit tests here?

crutkas:
Also should we move this logic out of UI and into core?

ralarcon:
What do you mean with storage.ide issue? Agree to check if we can add some unit tests. Apart from that, it is not strightfoward to move this to other place as we only can catch the "escape / cancel" action thru handling the _WizardBackoutException_ which occurs at UI level. 

We already have some methods which delete folders but this concrete combination do not looks to much reusable (IMHO)...

crutkas:
storage.ide was mentioned in https://github.com/Microsoft/WindowsTemplateStudio/issues/967

Issue #552, opened 2017-07-01
.NET 4.7


crutkas:
Just set up a new machine for testing and saw that we are 4.6.2.  Any reason to not just got to 4.7?  Needed to download new targeting packages.

crutkas:
VS solution is 4.5.2.  .net

crutkas:
punting to 1.4 due to issue @ralarcon raised in PR.  lets reduce complexity in this release

ralarcon:
I'm going to double check if adding the dependency to the installer solve the issue and what is the behaviour when users do not have installed W10 Creators Update (where the .NET 4.7 is being shipped).

sibille:
Verified in dev

PR #1027, opened 2017-08-30
Issue552 .net4.7

crutkas:
## PR checklist

Quick summary of changes

#552
Nope
not due to the CI tests


crutkas:
@ralarcon is this an easy item to fix or should we punt to 1.4?  I feel like just punt to 1.4 as we have a lot of additional moving parts

ralarcon:
Adding the dependency to .NET 4.7 cause the extension not being installed in machines without this dependency.

image

Then you got the error regarding the fx missing:
image

The log states that the fx version range and ends with the error:
```
...
SupportedFrameworkVersionRange : [4.7,)
....
Install Error : Microsoft.VisualStudio.ExtensionManager.MissingTargetFrameworkException: The extension 'WTS (vs2017 local)' requires a version of the .NET Framework that is not installed.
```
I notice that the machine I been working on yesterday had the .NET Fx updated automatically. Considering it is from the Creators Update, I guess there should not be too much target machines without the .NET 4.7 available. So in my opninion it is safe to include this change for this version.

I will update the PR adding the pre-requisites.

mrlacey:
>Considering it is from the Creators Update, I guess there should not be too much target machines without the .NET 4.7 available. So in my opninion it is safe to include this change for this version.

@crutkas do you have access to any telemetry on the versions of Windows & VS people use for UWP dev? Would be good to base this on facts if we can.
We also just did a lot to maintain support for versions of VS before v15.3.  Are we now saying that wasn't as important as we now require .Net Framework 4.7 to be installed?

crutkas:
asking

ralarcon:
Anyway, .NET 4.7 missing can be installed by the user, the log is clear... I think it is not an stopper.

crutkas:
yeah, i think we could just wait for 15.4 for this, no clear reason other than it would be a nice to have.

Issue #921, opened 2017-08-10
Update Settings docs to reflect changes to app theme selection.


mrlacey:
Update pictures and instructions to reflect changes made for #350

X-Ref #918 

Take special note of Pivot & CodeBehind

mrlacey:
closing as now merged to /dev

PR #1046, opened 2017-08-30
Updated Settings docs to reflect other changes made for v1.3

mrlacey:
- Which issue does this PR relate to? #921

Just docs changes.


templates/Pages/Settings.CodeBehind/Views/SettingsPagePage.xaml.cs
```diff
@@ -8,7 +8,7 @@ namespace Param_ItemNamespace.Views
 {
     public sealed partial class SettingsPagePage : Page, System.ComponentModel.INotifyPropertyChanged
     {
-        //// TODO WTS: Add other settings as necessary. For help see https://github.com/Microsoft/WindowsTemplateStudio/blob/master/docs/pages/settings.md
+        //// TODO WTS: Add other settings as necessary. For help see https://github.com/Microsoft/WindowsTemplateStudio/blob/master/docs/pages/settings-codebehind.md
```
crutkas: mvvm and mvvmlight don't have this todo?

crutkas: @mrlacey, ping :)

ralarcon:
Merging... I will update the comment afterwards in MVVMBasic / MVVMLight as I verified it is missing.

templates/Pages/Settings.CodeBehind/Views/SettingsPagePage.xaml.cs
mrlacey: Yes, the link in the codebehind version of the template now points to the code behind specific version of the docs. This [exists in the dev branch](https://github.com/Microsoft/WindowsTemplateStudio/blob/dev/docs/pages/settings-codebehind.md) but hasn't yet made it to master. It will when this change goes public too.

I didn't change the MVVM Basic & MVVM Light versions and they still point to the [generic page](https://github.com/Microsoft/WindowsTemplateStudio/blob/dev/docs/pages/settings.md) (with links to framework specific versions) as the alternative would be to add two new composition templates and that felt like a lot more effort than was worth including for the sake of saving someone clicking a link.

mrlacey:
> Merging... I will update the comment afterwards in MVVMBasic / MVVMLight as I verified it is missing.

@ralarcon not missing, just slightly different.

templates/Pages/Settings.CodeBehind/Views/SettingsPagePage.xaml.cs
mrlacey: MVVM Basic & MVVM LIght have the TODO in the ViewModel as that's where the majority of changes are needed.

Issue #1045, opened 2017-08-30
Update UWP Community Toolkit to v2.0


mrlacey:
The UWP Community Toolkit has just released version 2.0 ([announcement](https://blogs.windows.com/buildingapps/2017/08/30/uwp-community-toolkit-v2-0/))
We need to update templates to use this new version.

crutkas:
Note: toolkit takes UWP NuGet package 5.3.3

crutkas:
Got this working, adds some compile time as both nuget packages need to be fetched.

![image](https://user-images.githubusercontent.com/1462282/29898941-685a6e56-8d9d-11e7-9c31-ecec99192c19.png)

crutkas:
@mvegaca, the PR is ready but we moved this to 1.4 due to the the need to update Microsoft.NETCore.UniversalWindowsPlatform and the added first time run time it would add on a new box

sibille:
Master Detail is referencing version 1.5

sibille:
Verified in dev-nightly:
Wizard: 0.14.17283.65001
Templates: 0.14.17283.1

PR #1047, opened 2017-08-30
Update for UWP Toolkit 2.0, UWP .NET 5.3.3

crutkas:
## PR checklist

Quick summary of changes

#1045
5.3.3 UWP had to be updated due to UWP Toolkit 2.0
yup


ralarcon:
I made a small test an using the Visual Studio 15.3 references by default 5.4. so, maybe we could think in update the csproj / vbproj to reference this (but this would impact people running other Visual Studio Viersions)
``` xml
    <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
      <Version>5.4.0</Version>
    </PackageReference>
```
I would like to verify if the Visual Studio 15.3 includes the 5.3.3 / 5.4 and the behavior on vanilla machines (without having compiled any app) to verifiy that the project creation time (generation + nuger.restore) does not have impact when run from Visual Studio 15.3 (when it runs from 15.0 it will, as the Universal Platform distributed is 5.3.2.

ralarcon:
Closed by error... :D

ralarcon:
Verified. Pointing to UniversalWindowsPlatform 5.3.3 causes a big delay restoring the packages because of the download of packages from nuget. I will check with 5.4 as it looks to be the default in Visual Studio 15.3 (so I guess it is delivered as part of it).

ralarcon:
Using 5.4, and creating a BLANK app (Universal) takes a while to be created (it put 2,5GB in the nuget chache). If the we open a WTS, it takes a while too to be opened too (and it adds about 1 GB more to the nuget cache)... I bet that creating a WTS app with the nuget cache empty will take about the same time using 5.3.3. than using 5.4. So we need to decide what we prefer... or maybe we can delay this for 1.4 and align it with the VS2017 15.4 (which I know will have more changes and improvements).

crutkas:
agreed.  @mrlacey, i don't know if i see a strong reason currently for us to upgrade to 2.0 until 15.4 rtms.

i'm fine with the 5.4 update but by the time 15.4 rolls around, I'm pretty sure the default package will be 6.0.0

ralarcon:
Build failure is from a test.

Project Templates currently configured to TargetVersion 10.0.15063.0 and NETCore.UniversalWindowsPlatform 5.4.0. When ready, those values will be upgraded to 10.0.16XXX.0 and 6.0.0

PR #1049, opened 2017-08-31
Avoid clean up generations when building solutions

ralarcon:
#957

### DIFF
--- a/code/test/Templates.Test/BaseGenAndBuildTests.cs
+++ b/code/test/Templates.Test/BaseGenAndBuildTests.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Templates.Test
             await _fixture.InitializeFixtureAsync(language, this);
         }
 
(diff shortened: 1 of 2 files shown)


mrlacey:
What's the benefit and consequence of not deleting the files at the end?
Does it not leave a lot of generated projects on the file system?
The naming also seems backward. If requesting "cleanGeneration" then it doesn't delete the folder afterward. If I don't want a clean generation then the folder is deleted.

ralarcon:
The thing is that, after splitting in two the "Generations" and the "Builds" integration test, what i really do for "Build" is to call the "Generation"+"Build". If the generation test always cleans the generated solution then the build fails. 

So, in summary: 
* If you are in a "generation" test, then,  defaults to "true" -> clean (there was a mistake in this PR -I miss to remove the negation, fixed in #1050). 
* If you are in a "build" test, you call "generation with clean=false" and + "build", which also cleans. No rubbish is left.

PR #1054, opened 2017-09-01
Adding missing comment

ralarcon:
#921

Adding missing comment. Related to PR #1046


mrlacey:
This change is pointless and wrong.
It's duplicating a TODO comment which already exists in the ViewModel. 
It makes more sense for it to be in the ViewModel as this file is only used in MVVM projects and the majority of the changes suggested by the linked comment go in the VM.

ralarcon:
@mrlacey, you are right... I did't realized... I will remove it.

PR #1083, opened 2017-09-06
Update selected item foreground on ShellNavigationItem

### LINKED ISSUE #1078: Broken NavPanel item highlighting in latest dev build
The highlighting of the text and icon in the Nav Menu of the currently displayed item has been broken.
In the below, the circled item should use the accent color, like the other version in the image.

image

Above created with a MVVM Basic project

    <genTemplate:Item Name="wizardVersion" Version="0.14.17247.65001" />
    <genTemplate:Item Name="templatesVersion" Version="0.14.17248.13" />

mvegaca:
Fix #1078

### DIFF
--- a/templates/_composition/CodeBehind/Project.SplitView/Views/ShellNavigationItem.cs
+++ b/templates/_composition/CodeBehind/Project.SplitView/Views/ShellNavigationItem.cs
@@ -76,7 +76,7 @@ namespace wts.ItemName.Views
                 SelectedVis = value ? Visibility.Visible : Visibility.Collapsed;
 
                 SelectedForeground = IsSelected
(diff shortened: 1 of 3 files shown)


mrlacey:
Does this changing of System Theme names have issues with backward compatibility? i.e. does it work if running on a version of Windows older than the Creators Update?
If it's related to an underlying system resource change do we need to consider using different resources depending on whats available on the machine the app is running on?

crutkas:
@mrlacey, the change looked to be in https://github.com/Microsoft/WindowsTemplateStudio/pull/906/files and something i clearly did for some reason but honestly have the slightly clue why.

My gut says i did it to match what the navigation view will look like. https://docs.microsoft.com/en-us/windows/uwp/controls-and-patterns/navigationview

lets adjust back now and adjust later once we take a dependency on the navview.

PR #1088, opened 2017-09-06
Issue928 optimize vscommand load

ralarcon:
Added autoload only for solutions with projects as suggested by the VS team.

Removed generation context intialization in background. Now the is not executed until we are about to create a page or a feature (sessions registered in the telemetry will reflect users executing WTS actions).


crutkas:
this also looks to remove VB from menu.  We should create an issue to reenable it after we ship

code/src/Installer.2017/source.extension.vsixmanifest
```diff
@@ -27,6 +27,5 @@
         <Asset Type="Microsoft.VisualStudio.Assembly" d:Source="Project" d:ProjectName="UI" Path="|UI|" AssemblyName="|UI;AssemblyName|" />
         <Asset Type="Microsoft.VisualStudio.VsPackage" d:Source="Project" d:ProjectName="%CurrentProject%" Path="|%CurrentProject%;PkgdefProjectOutputGroup|" />
         <Asset Type="Microsoft.VisualStudio.ProjectTemplate" d:Source="Project" d:ProjectName="CSharp.UWP.VS2017.Solution" d:TargetPath="|CSharp.UWP.VS2017.Solution;TemplateProjectOutputGroup|" Path="ProjectTemplates" d:VsixSubPath="ProjectTemplates" />
-        <Asset Type="Microsoft.VisualStudio.ProjectTemplate" d:Source="Project" d:ProjectName="VBNet.UWP.VS2017.Solution" d:TargetPath="|VBNet.UWP.VS2017.Solution;TemplateProjectOutputGroup|" Path="ProjectTemplates" d:VsixSubPath="ProjectTemplates" />
```
mrlacey: This shouldn't have been there anyway. I deliberately took it out when the VisualBasic branch was merged. I will investigate.

ralarcon: Sorry, I added it some point in time... didn't show this comment till today (you would have saved some time).

Issue #1104, opened 2017-09-07
UriScheme feature breaks for MVVMLight if start page name is changed


mrlacey:
**Repro steps**

- New project, NavPanel & MVVMLight
- Change the name of the first/main/home page
- Include the UriSchemeFeature

**Expected Behavior**

- All compiles fine.

**Actual Behavior**

- Change in name of the default page isn't handled by the UriScheme templates and it tries to reference the VM for the default page name value which doesn't exist, so the code doesn't compile.

- `GenAllPagesAndFeaturesRandomNamesAsync` should have caught this!

_discovered while investigating #1097_

mrlacey:
@ralarcon In that there's a CI build done by AppVeyor, are the results of the tests done as part of the process publicly available?
- If so, can they be linked from the Build Status on the main readme?
- If not, can they be made public? so it's possible to verify the tests that are run as part of a build

ralarcon:
They are... you can Access from the PR checklist if not wrong, somehting like: https://ci.appveyor.com/project/ralarcon/windowstemplatestudio/build/CIBuild_0.14.79.dev/tests

mrlacey:
It seems the CI is running `GenerateAllPagesAndFeaturesAndCheckWithStyleCopAsync` but not `GenAllPagesAndFeaturesRandomNamesAsync`. Running both of these is the minimum we need to do to have confidence in the templates. Over the summer, before we had the AppVeyor instance we went down to running just these two sets of generation tests as they are the simplest way to cover the widest range of range of scenarios. To leave out either puts a lot of uncertainty over the quality.
That's a separate issue.

The reason this wasn't caught by the automated tests is that there's a bug in the GenAllPagesAndFeaturesRandomNamesAsync test. I'll raise a separate issue for that.

ralarcon:
@mrlacey I'll include `GenAllPagesAndFeaturesRandomNamesAsync` as part of the "Minimum" execution test. Open the issue for the bug in GenAllPagesAndFeaturesRandomNamesAsync and we will have a look on it tomorrow.

mrlacey:
@ralarcon actually don't.  You need to include `BuildAllPagesAndFeaturesRandomNamesAsync`.
[...]

PR #1107, opened 2017-09-07
Extend minimum test execution set

ralarcon:
Included the Generation+Build of all project combinations (Frameworks and Project Types) with all features and pages in the Minimum execution set. This will increase the generation+build tests surface to catch more issues within the CI

#1104

The build time for CI will be increased to run the minimum execution set (expected 10-15 mins now).

Yes

### DIFF
--- a/code/src/Core/ProgrammingLanguages.cs
+++ b/code/src/Core/ProgrammingLanguages.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Templates.Core
         public static IEnumerable<string> GetAllLanguages()
         {
             yield return ProgrammingLanguages.CSharp;
-            yield return ProgrammingLanguages.VisualBasic;  // This is currently disabled until full VB support is enabled
+            // yield return ProgrammingLanguages.VisualBasic;  // This is currently disabled until full VB support is enabled
         }
     }
 }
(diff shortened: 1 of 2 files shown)


ralarcon:
Not sure why it fails in the AppVeyor instance... locally the tests are working... tomorrow will investigate further.

mrlacey:
VB tests don't yet need to be run as part of the minimum set

mrlacey:
> VB tests don't yet need to be run as part of the minimum set.

From **Microsoft.Templates.Core.ProgrammingLanguages** in the dev branch
```
public static IEnumerable<string> GetAllLanguages()
{
    yield return ProgrammingLanguages.CSharp;
    yield return ProgrammingLanguages.VisualBasic;  // This is currently disabled until full VB support is enabled
}
```

The above was created as part of the preparation work for VB. The second yield line was commented out when the VB work was merged and the comment line made sense (it's not disabled when uncommented) as there was nothing that should be being done with the VB templates for v1.3.

Because this line has been uncommented the automated tests are doing, potentially, twice as much work as they _currently_ need to.

This is now the second thing related to VB functionality that has been added to the executing code when it shouldn't have.
EIther someone is enabling this as part of something else and not explaining or understanding the consequences. 
Or lots of changes are being incorporated into PRs that shouldn't be.
With so many tests not working as they should or disabled, this leaves me very concerned about the state of the code in general, and even more on the approach to a release.

ralarcon:
@mrlacey, it was me as part of the other change... I've included this to run the VB projects as I enabled the VB in the installer... I thought it was needed :D

Issue #1112, opened 2017-09-08
Ranaming Settings page cause page not being created properly


ralarcon:
### Repro steps
Create project, add a Settings page and rename the default name i.e to Settings123. The setting page is generated wrong:

![image](https://user-images.githubusercontent.com/2638875/30207990-eafec302-9491-11e7-838e-08df664888fa.png)

Wizard: 0.8.17251.1
Templates: 0.8.17251.1

The issue is related to the template where the "replacement name" is not set properly (in part because it is very similar to the default name):

<data name="**SettingsPage**_Title.Text" xml:space="preserve">
    <value>SettingsPage</value>
    <comment>Page title for SettingsPage</comment>
  </data>
  <data name="SettingsPage_Theme.Text" xml:space="preserve">
    <value>Choose Theme</value>
    <comment>Choose theme text for SettingsPage</comment>
  </data>
  <data name="**Settings**_Theme_Dark.Content" xml:space="preserve">
    <value>Dark</value>
    <comment>Dark theme text for SettingsPage</comment>
  </data>
  <data name="**Settings**_Theme_Default.Content" xml:space="preserve">
    <value>Windows default</value>
    <comment>Windows default theme text for SettingsPage</comment>
  </data>
  <data name="**Settings**_Theme_Light.Content" xml:space="preserve">
    <value>Light</value>
    <comment>Light theme text for SettingsPage</comment>
  </data>
 
It is a subtle change which suggest that we should think in define the replacement names in the templates more obvious to avoid this kind of errors.

sibille:
Verified on pre-release 
Wizard: 0.8.17251.2
Templates: 0.8.17251.2

PR #1116, opened 2017-09-08
Use SourceName for Settings page ressources

sibille:
Quick summary of changes
Using SettingsPage on resources and xaml to support renaming of the page correctly
#1112


crutkas:
good catch.  Is this something we could have caught with some type of test?

Issue #1037, opened 2017-08-30
Create tool for analyzing/verifying templates


mrlacey:
With more people creating templates and the expectation (and hope) that this will continue, it will be useful to have a tool that can do some level of automated analysis of generated templates.

The aim here is to spot any mistakes, errors, or potential issues quickly and provide some confidence that the template is correct.

Not a substitute to manual checking and other testing of templates but a compliment to it.

This should also integrate with the coded tests that we currently also have but also be able to be run separately.

X-Ref #919

I suspect that any such tool will be added to over time but having something to get us started would be a bonus.

ralarcon:
We had in mind an extension for Visual Studio Code to allow users authoring templates to make checks... but nothing started yet.

crutkas:
great ideas @ralarcon and @mrlacey!

mrlacey:
I want something scriptable for automated tests as well as well as stand alone and tied to the core code base as much as possible to avoid issues with maintaining concurrent versions.
A VSCode extension may be useful for some in creating templates but I'm aiming for something more broadly applicable.
I already have my own tool for validating JSON that I'll make public and use as part of this.

ralarcon:
Good approach. The extension for the VSCode could be implemented "on-top", isn't it?

mrlacey:
Also, need to update https://github.com/Microsoft/WindowsTemplateStudio/blob/dev/docs/templates.md

mrlacey:
For the initial version, I'm not planning on supporting localized output.
As this isn't part of the wizard and won't be included in anything that's shipped I'm assuming this won't be an issue. 
If translations must be supported I'm hoping this can be added in a later version.
Adding this in at this stage will delay the creation of this tool. I think it's more important to see if others find it useful before spending time on globalization.

mrlacey:
For the initial version, I'm planning the following:
- Ensure that template.json contains only valid properties
- Ensure the values specified for properties in template.json (will catch typos, etc.)
- Ensure files referenced in template.json exist
- Ensure all Tag values are valid/appropriate
- Ensure naming standards for VB templates are followed
- Ensure all defined symbols are known/supported
- Check for unused files in the template
- Check for uniqueness of template identities
- Check that defined dependencies exist
- Check for identical files in multiple templates (that could be combined)

Intended to be easy to extend with other checks as needed in the future.

[...]

PR #1124, opened 2017-09-11
Issue1037 template validator

mrlacey:
For #1037 ## PR checklist

- Adds a new tool to validate template files.
- Also, runs the tools tests as part of the automated tests.
- Included some changes to the templates for issues found during the creation of the tool.
- Includes appropriate addition to the template docs.

The original intention behind creating this tool was to help me when creating templates but I also hope it will be useful to others. If not found to be used/useful it can be re-evaluated at a later time. If wider use is adopted I'd like to add some unit tests to ensure it continues to function as intended. I don't feel spending time on such tests at the moment would not be the best use of my time.

Also, introduces a new external dependency but that also uses the MIT license and isn't included in anything that is installed.


ralarcon:
@mrlacey, can you please add the "Analyze" configuration for the `TemplateValidator.csproj`, it is just a copy but in this configuration the "analisys" is enabled at build time by default and causes build to fail if there is a rule violation. This is the configuration we use in the CI to ensure the StyleCop rules being verified.

mrlacey:
@ralarcon It' looks like it's the "Any CPU" vs "AnyCPU" issue where the Solution version doesn't match the project version but I can't see how this is different from other projects in the solution. Can you see what's wrong?

ralarcon:
Yes, let me have a look on it...

ralarcon:
@mrlacey... your change was right and the Big.sln builds perfectly... the thing is that the solution used to run the CI in AppVeyor is "AppVeyorCI.sln", so is in this solution where we need to add the project and the new configuration (Analyze) for the TemplatesValidator project. 

I needed to create the AppVeyorCI.sln solution since the AppVeyor build server was not able to build the Installer.2017 project (it does not have installed the Visual Studio Extension capabilities). What I do not understand is why the TemplateValidator project is being build because it is not in the solution itself (maybe because it was doing a nuget restore for the big.sln)

mrlacey:
Yep, added to the AppVeyorCI.sln and all works. 
My guess is that it found the project and tried to build it because `Templates.Test` has a reference to it so it can run automatically all the tests against the templates we ship.

Issue #425, opened 2017-05-31
What is a good way to add a top toolbar (along with left navbar) to a page or to all pages?


dragansr:
Adding CommandBar element in Page.TopAppBar works on individual pages, but it hides main "hamburger menu"

Is there any sample project where a toolbar is integrated in the templates? 

Eventually it would be nice to have option for a toolbar / CommandBar in the wizard also.

mrlacey:
@dragansr despite its name `Page.TopAppBar` puts the `CommandBar` at the top of the window, not the `Page` that declares it. Because the NavigationPanel works by putting a page within the ShellPage this is why the CommandBar appears over the hamburger control.
You can avoid this by creating the creating the CommandBar inline (within the content of the page) and then it won't appear over controls in the ShellPage. See the code at the bottom for an example.

At the moment there's nothing in the generated templates that uses a CommandBar.
The CommandBar is a very broadly used control, integrating it with the WTS templates raises lots of potential questions:
- How should this be included?
- Does it need a special page type or to be added to existing pages? 
- If existing pages, should it be any/all or just some?
- If just some, which?
- Add to the top, bottom, both, or inline? Or should it vary based on page, project type, and/or device family?
- Include primary and/or secondary items?
- Just regular buttons or also include toggle buttons, separators, or anything else? (context menus?)

Having lots of options is bad for templates as it typically means adding something that won't be exactly what everyone needs and so they have to remove it. Or it means providing a lot of configurable options which can be hard for people to understand and know how to use. This is particularly important for a tool like WTS as I don't expect it to be used lots (very few people build lots of apps regularly) and so usage and options need to be very clear.
This doesn't mean ruling it out, just that need to be very careful with what's added and why.

&nbsp;
&nbsp;
&nbsp;
&nbsp;
&nbsp;
The code below shows an inline CommandBar added to a Blank page.
```
<Page
    x:Class="App52.Views.BlankPage"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    mc:Ignorable="d">
    <Grid
        x:Name="ContentArea"
        Margin="12,0,12,0">

        <Grid.RowDefinitions>
[...]

PR #1130, opened 2017-09-11
NavPanel + CommandBar docs

mrlacey:
Added documentation explaining issues and how-to's for showing a
CommandBar in an app using the NavPanel project type so the bar and
hamburger menu don't overlap.
#425


hupo376787:
I have encounter the same problem.

mrlacey:
> I have encounter the same problem.

@hupo376787 can you explain what you mean by this? Is it a problem with the docs in the PR or are you just confirming that the issue the doc change covers exists?

mrlacey:
I don't really want to add more comments to the generated code that just says "Hey! there's some docs about X over here" as if we do that for all docs then we would (hopefully-as more docs are created) end up with these comments all over the generated code.
Having the comment and link about changing the icons make sense because everyone will have to change the generated code but this will only impact some people.

Issue #965, opened 2017-08-17
Have a generator test that runs store validation/WACK against the generated project


mrlacey:
X-Ref #907 

We should be totally confident that the generated project will get to the store and that we don't do anything that could prevent this.
This should include catching anything related to 3rd party libraries we add references to.

Not even sure how to do this (or if possible--as haven't looked) as I expect this will require associating with a dev center app entry. Will require some investigation but I think this is definitely worth having.

crutkas:
great idea

mrlacey:
Doing now as I want to run this against the VB generated projects to verify their completeness.

mrlacey:
Closing as merged into dev. It isn't part of wizard or templates and so doesn't need to wait for a build to verify it.

PR #1147, opened 2017-09-15
Run WACK tests against generated projects

mrlacey:
For #965
Build Release version of projects generated with all pages and features, then generate an AppxBundle from that and then test the appxbundle with the Windows App Cert Kit (WACK) which includes store
readiness/appropriateness checks.
The test also replaces default images (with WTS logos) or these would cause the WACK test to fail.

These tests are very slow to run and need to run under administrator privileges or the WACK test will raise UAC prompts.

Tests are currently part of Templates.Tests but may need to move them out if they cause an issue when running all tests.
These tests cannot be run on the CI server.
Not only do they require UAC they are also **VERY** slow (20+mins per test for me. This is machine dependent but primarily due to the .Net Native compilation for a release build. This time is then multiplied by M*N projects.)

Because they're so long running I've set a "ManualOnly" execution trait and set the Skip property on the `Theory` in the hope that this means they won't be picked up by any automated testing. If that's not adequate or sufficient I'm open to suggestions for how best to have such tests as part of the current codebase.

+ Fixed tests (& VSEmulator) adding PFX files to project with the wrong build action as this caused the WACK tests to fail. (Should be "None", was "Content". FakeMsBuildProject was doing something different to actual Visual Studio.)

**Seriously** this takes _ages_ to run
image
I suggest running them over night or taking a reallllly long coffee break as the WACK tests mean you can't use the machine while they're running.


ralarcon:
@mrlacey we can have an specific build to be run daily within a specific build server with the pre-requirements needed. No problem at all. I've created the issue #1148 pointing to release 1.4 (considering low priority we will tray to include it) to track this.

Issue #607, opened 2017-07-05
Wizard: Break features and pages out


crutkas:
We've hit that time now where we need to break pages and features out.  Too many options for a single page IMO.

mrlacey:
split over two steps in the wizard (getting to 3 in total)?
Pages, then features seems sensible to me.

crutkas:
Yes.  Navigation & Framework -> Pages -> Features.

Once xamarin comes into play, we'll need maybe to rethink the first step as step 1 then would be Platforms

ralarcon:
Do you think this can wait until 1.2 is delivered?

crutkas:
I marked it for 1.3 when i created it.  Just wanted to create the work item.  Also call out the fact we'll need to think about the xamarin adjustment for platform :)

mrlacey:
A good candidate to bump to 1.4 as only added 1 feature and no pages in 1.3 - @crutkas?

crutkas:
agreed.  lets push to 1.4 but IMO this will be needed for xam + wizard update

mvegaca:
I'm thinking about start working on it tomorrow morning.

mrlacey:
Should the splitting of ProjectType and Framework (& the XF option) be added at the same time? They seem very closely related and if major changes will need to be made for this any such changes should also consider what will be needed for that. I can't find an issue for that work but I know it's been discussed. Making sure this is fully understood will be necessary before any work can begin.

There's also a potential overlap with #875 and #827 that should probably also be considered at the same time even if not implemented together.

mvegaca:
I think that project type and framework should keep together in step 1.
I think that re-categorize features are only related to the concept the templates, categories are composed using a static field on templates json.
Redesign adding new pages and features can be a different task, my idea was to work on this keeping the same code and UX

mrlacey:
/cc @crutkas Re: "I think that project type and framework should keep together in step 1." I know there has been lots of discussion on splitting this up. Partly for the XF work and partly to better support project types that aren't supported by some frameworks (which I know was being discussed for Prism). If changes are being made to the wizard they should consider the other planned changes too so anything done now doesn't make things harder for the known future changes if this can be avoided.

If the redesign and UI changes are large and are also going to happen for the next release then doing them separately will likely lead to unnecessary and wasted effort spent making changes and keeping the current UI. Whether this is an issue depends on how much time is available to be spent on this and how big the planned changes are.

[...]

PR #1169, opened 2017-09-19
Include tests to validate project config info query / infer

ralarcon:
Added validations to verify correct use of ProjectConfigInfo

#607

Nop.

Yes.

### DIFF
--- a/code/test/Templates.Test/GenTests/GenProjectsTests.cs
+++ b/code/test/Templates.Test/GenTests/GenProjectsTests.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Text;
 
 using Microsoft.Templates.Core;
 using Microsoft.Templates.Core.Gen;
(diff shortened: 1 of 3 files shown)


code/test/Templates.Test/GenTests/GenProjectsTests.cs
```diff
@@ -48,6 +49,59 @@ public GenProjectTests(GenerationFixture fixture)
 
         [Theory]
         [MemberData("GetProjectTemplatesForGenerationAsync")]
+        [Trait("Type", "GenerationProjects")]
+        public async Task GenEmptyProjectCorrectInfeProjectConfigInfoAsync(string projectType, string framework, string language)
```
mrlacey: What does this test do? The name is not clear and appears to contain a typo.

ralarcon: When a solution is generated we add some metadata to the manifest to identify the framework and project type. If this information is missing, we try to infer it (there is a typo in the method name it should be "Infer" instead of "Infe") and add the correct tags to the manifest. This is done by the ProjectConfigInfo class.

Now, every generation test ensures we are adding the expected tags adding [a new assertion](https://github.com/Microsoft/WindowsTemplateStudio/pull/1169/files/77d09d84e02866747761799174745127c09feebd#diff-4b6d07657a0d2ccdc4eec08d69b11e50R58), that is validating the information generated matches with what is expected. 

In particular, this tests forces to make the infering removing the generated tags and calling the ProjectConfigInfo class. Then validates that the re-generated tags are as expected. This was implemented to have backward compatibility with project generated before right-click exists. And it is true that maybe there should not be too much projects created without the metatada, but just in case, we think it is interesting to validate that the inferention algorithm works in every case.

mrlacey:
Why wasn't the typo fixed?

sibille:
It's on its way, coming in the next PR

Issue #1164, opened 2017-09-19
Image gallery page


mvegaca:
The idea here is to create a new image gallery page template that uses simple navigation RS2 and connected animations on RS3.

mrlacey:
Will need suitably sourced images to include in the generated app. Discussions about sourcing and including images in the past came to no definitive conclusion. What remote source and/or what can be included in the generated project. Curious about what was decided for this

ralarcon:
We talked to generate sample images using a template with a number variation ("1", "2", "3", "4"...) and posibly a color variation (first approach will be b&w)...

mvegaca:
I've created a template that generate default images using gray colors for having a great aspect on dark and light themes.


![samplephoto4](https://user-images.githubusercontent.com/5556784/30591006-61aacb7e-9d41-11e7-9a64-bb2384829787.png)

mvegaca:
Attached mockup version for RS2
[PhotosApp.zip](https://github.com/Microsoft/WindowsTemplateStudio/files/1314664/PhotosApp.zip)

Gallery page will add:
- Gallery page
- Photo detail page
- Nuget reference to Microsoft Toolkit UWP
- SampleDataService dependency
- Sample images on Assets
- String Resources

Things in mockup:
- Compleated gallery page compatible with dark and light themes.
- Accessibility checks completed.

Things TODO:
- Add FCU connected animations to mockup
- Create a template from mockup
- Create composition templates for different frameworks and project types
- Add conditional code for rs2 and rs3 generation.

Persona notes:
- I've added UWP toolkit for adding ImageEx control, I think it could be great on this page to work with big images on final applications.
- I've not added AdaptativeGridView because it adapts the images increasing only the with and does not respect the aspect ratio.

Application preview

![image](https://user-images.githubusercontent.com/5556784/30599259-349a9b3a-9d5c-11e7-82c4-e5d022dc4da1.png)

mvegaca:
Working on template on branch [Issue1164-gallery](https://github.com/mvegaca/WindowsTemplateStudio/tree/Issue1164-gallery). Ready to try and generate apps but not ready to be merge by now.

mvegaca:
Template finished for C#: MVVMBasic, MVVMLight and CodeBehind frameworks on #1177

Pending translation to CaliburnMicro framework and Visual Basic language.

mvegaca:
Caliburn Micro translation request in #1200

sibille:
[...]

PR #1177, opened 2017-09-21
Image Gallery Page

mvegaca:
Issuse #1164

@crutkas can you review descriptions and add the path icon?


mvegaca:
Finish, it can be approved

Issue #831, opened 2017-07-19
Need documentation on when and where each type of automated test is run


mrlacey:
To avoid issues with builds timing out and to ensure that tests are run appropriately, We need clear documentation on which tests are run when.

This information isn't currently visible publicly as there's no access to the build machine.

We need to be sure that new tests are grouped/classified appropriately.

X-Ref #828

PR #1193, opened 2017-09-26
Added documentation on which tests are run in which build

sibille:
Quick summary of changes
Added documentation on which tests are run in which build
#831


mrlacey:
This is a good start but it also needs to be clearer about which build relates to which column on the home page. Is "Templates.Test.Gen" the same as "Templates Publishing"?

image

Is it even true that a build passing includes all related tests passing?

Related: it would also be good if the results on the readme page indicated when they were last run and against which branch.

sibille:
Good point, I'll adjust the names in the documentation to what we use in the Build Status table. 
Yes, failing tests will make build fail.

ralarcon:
@mrlacey, you shoud check the dev branch page (not merged to master yet).

mrlacey:
@ralarcon not everyone who looks at the project will know to look to the dev branch to see the latest.
This also highlights that changes to this doc should be applied to master more frequently too.

Issue #196, opened 2017-04-18
Feature: Launch Feedback Hub


mrlacey:
Add the ability to launch the feedback hub to gather user feedback.

Possible functionality:
* On Blank and Pivot&Tabs projects, add to the settings page
* On NavPanel projects add as a link in the SecondaryItems (bottom) menu

Making this a feature, not just an addition to the Settings page as it introduces a new dependency (Microsoft Store Services SDK)

For reference: https://docs.microsoft.com/en-us/windows/uwp/monetize/launch-feedback-hub-from-your-app

mrlacey:
I have a PoC of this and have shown it to Clint & Michael. Will add as a branch next week when I get back from build.

crutkas:
https://docs.microsoft.com/en-us/windows/uwp/monetize/microsoft-store-services-sdk#install-via-nuget for nuget install.

I think we should discuss event logging for VSMC and Store SDK.

Love the ads idea

crutkas:
Should this take a dependency on the settings page?  I feel like yes.

mrlacey:
@crutkas why a dependency on the settings page? Are you thinking it should be something added there? (rather than in the TitleBar?)

crutkas:
My thoughts are what app(s) do that?

mrlacey:
> My thoughts are what app(s) do that?

@crutkas which "that" are you referring to?

crutkas:
doing the feedback hub launch in the titlebar

mrlacey:
There are lots of places this could go:

- Add to the title bar (like Visual Studio)
- Add to the Settings page (like Groove)
- Add to primary navigation menu items. (like the News app)
- Add to the secondary navigation items (above settings)
- Another common place for such a link is an "About Page" (like Calculator) but we don't have one of those.
- As secondary item (hidden behind ellipses) in a command bar at the top of the page (Like Edge & Movies&TV)

Does it have to be in the same place for every project type? I don't think it does.

My proposal for this is:
- If a SplitView project add to the secondary menu (above Settings - I think this highlights how it's different from other menu options which are about navigation within the app)
- If a Pivot&Tabs add to the Command bar at the top of the page. This will, of course, also require adding the ability to add such a command bar (as a hidden feature.)
- Do not include in a blank project. With a blank project, the responsibility for the UI is entirely down to the developer and so they can choose to add a feedback hub link wherever they choose.

Does anyone else have an opinion or preference?

[...]

PR #1214, opened 2017-10-02
Issue196 feedback hub

mrlacey:
For #196
New feature adds feedback hub link to settings page
+ enables pages to be dependencies as well as features
+ enables chained dependencies
+ adds FileNameSearchPostAction to support merging with a file of unknown name
+ tests for tags starting "wts.fileNameSearch" in TemplateVerifier
+ docs updated for "wts.fileNameSearch[0-9]"
+ updated ApiAnalysis ref in TemplateVerifier to get needed bug fix

Does not add right-click support as doing so will require the ability to detect the settings page being installed in an app with a different name


sibille:
Matt, I think when we generate a feature as part of a "new project generation", there is another way to resolve the dependencies name (the name the user chose) and pass it by parameter to the template. 
Template engine now allows to participate parameters in file renaming, specifying the parameter like this:  https://github.com/dotnet/templating/pull/444/commits/2c5d851f20a3bca150e02f714bcf772de660c014
```
"symbols": {
    "wts.rootNamespace": {
      "type": "parameter",
      "replaces": "Param_RootNamespace"
    },
    "wts.itemNamespace": {
      "type": "parameter",
      "replaces": "Param_ItemNamespace"
    },
    "baseclass": {
      "type": "parameter",
      "replaces": "System.ComponentModel.INotifyPropertyChanged"
    },
    "setter": {
      "type": "parameter",
      "replaces": "Set"
    },
    "dependencyName": {
      "type": "parameter",
      "replaces": "DependencyFile",
      "fileRename": "DependencyFile" 
    }
  }
```
The last one would contain the name the dependencies name. We'll have to resolve this name when adding templates to the generation queue. Having that parameter, you can generate a "classsic" postaction I think, which would be more straightforward and robust. 

What do you think? Is there anything I didn't see?

mrlacey:
@sibille what the templating library added with **fileRename** appears to be a limited version of what WTS could already do with SourceName. I say limited because it doesn't do partial file renaming. File renaming such as for file suffixes like *ViewModel, *.Page, & *_postaction.

If we attempted to use the templating filerename feature to handle unknown file names we'd then have two methods of handling the changing of file names but working slightly differently.
We'd still have to do work to identify what needs to change and handle file suffixes too. The file search approach I've implemented is how I'd hoped to handle right-click support here too, so we'd need a different approach for that too.

I don't think trying to switch to the filerenaming functionality for this feature is appropriate. If we adopted it as the alternative for everything we're doing for renaming then that would be different.


This also highlights the need to monitor changes to the underlying templating library and how we manage the co-development of features or other things they've added that might be useful if we knew about them.

sibille:
@mrlacey. I don't think it's a limited version of template engine's sourceName.
FileRenaming supports renaming of partial filenames. The parameter dependencyName from the example replaces "DependencyFile" in any file that is part of the template as well as in file names. 

We already use parameters for namespaces and homepagename, so it's nothing new, it's just a parameter that allows replacements in fileNames too.

This functionality is relatively new, but was implemented as part of this issue https://github.com/dotnet/templating/issues/410 we opened time ago about having the posibility to have two source names.

You're right that this will not solve  the right click situation, but I think it's easier to use and less error-prone.

ralarcon:
If we think this feature is mandatory for rigth-click then it is crucial to be able to solve it. I like the power from the "search" approach but at the same time it looks less robust and probably more difficult to understand (but this is not an issue... we have a lot of things difficult to understand).
Does it make sense to have a combination of both approaches?

mrlacey:
I'd really argue against supporting both ways of handling file renaming. It is too likely to create confusion about what's possible or how things should be done. If it was possible to use either way we'd need to add testing for both and other dev and maintenance efforts would increase.

Either we plan to switch to the built-in templating method or we don't use it at all. Using it in a different way just so we don't need the search postaction seems like the worst option as it adds: more work now and in the future; possible confusion over how to handle renaming in other new features; confusion about using filerename for something other than it was truly intended.

Not switching to it risks setting a dangerous precedent for not using the built-in templating functionality. Are we happy being similar but different in multiple ways? Forever?

Does the need we currently have to rename before a postaction merge occurs prevent us using the built-in filerename functionality?

If going to switch it should be done as a high priority (i.e. in milestone 1.5) as the longer it takes to implement the more work it will be as it will impact more other code.

If we're going to switch to the built-in filerename functionality then it makes sense to wait and implement this using that. We'll have a list of files that have been renamed and so not need the search capability. I can close this PR and resubmit when it is updated to support new core functionality. Even if we don't switch how we handle renaming.

I'm not happy moving to the built-in renaming functionality for this release. It's potentially a lot of work and could be quite far-reaching in terms of impact on the codebase.  If we're going to need to do this, the feature should be retargeted to the next milestone.

I think it's worth removing the right-click functionality from this discussion for now as it just risks confusing the debate about how to handle file renaming.
However, I don't think anything needs mandatory right-click support. In an ideal world, everything would have it but it's not a requirement. This priority may change once any app can be converted to support right-click add though, as it will increase demand for the capability.
We can easily recreate a version of the search postaction functionality if needed for future right-click support.

@sibille  have you got the filerename functionality working? I can't get it to work from the source and the related unit tests fail when I run them.

Through looking through the templating source code I've also discovered the kind parameter `file_contains` which looks to do what I've done with the searchpostaction. If I can get this working I think we should switch to using that instead. Based on this discovery, I think we should cancel this PR.

I'm going to leave this open while we decide about what to do with file renaming.

sibille:
@mrlacey, I got a bit lost, let's go step by step. ;-)
SourceName and  parameters (with or without filerenaming) are both out-of-the box functionalities of template engine. We already use both inside of the templates. Blank page uses as sourceName "BlankView" and parameters like "Param_RootNamespace", both values cause replacements (done by template engine) in the template files. The only new thing is that now parameters can be used to replace parts of filenames too, we did not use this earlier because it was not there. 

I'll prepare a branch with an example of using a parameter to replace the name of a dependency in a postaction file, to show you better what I'm refering to.

I'll have a look at the file_contains too, maybe thats our solution.

sibille:
@mrlacey you can see the PoC here https://github.com/sibille/WindowsTemplateStudio/tree/poc_parameterfilerename

I added a TestFeature that depends on SettingsPage and adds a comment to the SettingsPageViewModel using a postaction. 

I added a parameter to the TestFeature that holds the name of the dependency template. To make it generic the parameter's name is the same as the dependency's name. 
```
    "wts.Page.Settings": {
      "type": "parameter",
      "replaces": "Param_SettingsPageName",
      "fileRename": "Param_SettingsPageName" 
    }
```
I added [this](https://github.com/sibille/WindowsTemplateStudio/blob/39e608c85b321b6f276eef26352f56d34c9c2b5a/code/src/UI/Generation/GenComposer.cs#L176) code to the GenComposer to inform the name of the dependency parameter if its there. 
I think it's a good thing for a template to know the name of it's dependency. (In other cases we did not need this as dependency names never changed, but maybe we should include this there too, although that would be a 1.5 issue.) 

Apart from this I included the changes you did to be able to depend on a page and allow double dependencies.

For the right click I think @ralarcon was refering to use your search mechanism to find out the name of the dependency and then inform it with the dependency parameter.   

I would have loved to have checked this option earlier, but we did not make the connection between those parameters and renamed dependencies until looking at your PR the other day and I wasn't aware also that they had implemented it yet.

ralarcon:
We are including some changes related to the rules for StyleCop, we expect to do not impact you as those are going to be done over the Wizard.

mrlacey:
Updated this to make it much simpler, based on the PC from @sibille

mrlacey:
checks failed as dev has errors causing CI failure

ralarcon:
@mrlacey there are some tests from the "UI" which are failing after fixed the other errors... it looks that is something related to the MVVM Light license...

_Microsoft.Templates.UI.GenException : Dependency template wts.Page.Settings.MVVMLight not found for framework MVVMLight._

Not sure way. @mvegaca can help out having a look on it.

mrlacey:
The dependency got renamed and it wasn't flagged when I merged the latest from the origin. As the code in dev was failing on the CI I was waiting for that to be fixed otherwise I was just continuing to merge with something that was known to be broken.

Issue #1184, opened 2017-09-22
Adapt CaliburnMicro's Master-Detail page to new Master-Detail design


sibille:
We made enhancements to the master-detail page to adjusts its design to master detail redlines. 
CaliburnMicro is still using the old design, we should adapt to the new one.

sibille:
Caliburn.Micro's Master Detail designs updated, pending revision in #1183. 
Caliburn.Micro uses control from UWP Community Toolkit, should we change too? I remember we had a discussion on this, but unsure about the outcome.

mrlacey:
Can you find the past discussion? I think we didn't want to rebuild for the sake of it.
We should use the same for all frameworks.

sibille:
Found it!! https://github.com/Microsoft/WindowsTemplateStudio/issues/157

mrlacey:
It seems there wasn't a good enough reason to make the move previously. Have things changed? What are the implications for changing now?

ralarcon:
The thing is that the Caliburn Master / Detail template is using the toolkit, so we need to align in one sense or the other.

mvegaca:
I vote for using toolkit control in all frameworks.

mrlacey:
This isn't just a straight-up vote. justify decisions either way.

mvegaca:
If  we are now using toolkit on SplitView, could be a good idea to use it on MasterDetail and reduce the generated code.

crutkas:
My fear with that control is we abstract away too much. We had that issue open for a long time and no community member really pushed for it.  

At this point also, why would we shift.  Does the new image gallery style page use this control? 

The reason why we took the split view was so we could have an easy transition to Nav view control once the next update of Windows is released post Fall Creators Update.

ralarcon:
The image gallery do not use the toolkit control as we didn't see any advantage on using it.

From my point of view we can live with the fact of having the Caliburn.Micro template based on the toolkit and the others not. At the same time, I don't now how mandatory is to have all the templates really similar (as much as possible), considering Caliburn has more differentiations than the others.

I think we can leave the things as they are right now (refactorizing the sytles and improving what is needed).

crutkas:
Wait wait wait.  All code generated should be basically the same across frameworks and project types. 

Why is one master detail not like the others

ralarcon:
Basically because Caliburn.Micro master-detail template use the toolkit. We didn't found the differencies but understood the templates implementation were already checked.

[...]

PR #1227, opened 2017-10-04
Fix ActiveItem initialization in Caliburn.Micro

mvegaca:
Issue #1184

I've fixed ActiveItem initialization on Caliburn.Micro.
Could @nigel-sampson please review that? I don't know if this is the best approach to do that on Caliburn.Micro Framework.

### DIFF
--- a/templates/Pages/MasterDetail.CaliburnMicro/Views/MasterDetailViewPage.xaml
+++ b/templates/Pages/MasterDetail.CaliburnMicro/Views/MasterDetailViewPage.xaml
@@ -38,7 +38,7 @@
             Grid.Row="1"
             x:Name="MasterDetailsViewControl"
             ItemsSource="{x:Bind ViewModel.Items}"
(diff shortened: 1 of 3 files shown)


mvegaca:
@nigel-sampson when possible can you have a look here?

templates/Pages/MasterDetail.CaliburnMicro/Views/MasterDetailViewPage.xaml
```diff
@@ -38,10 +38,11 @@
             Grid.Row="1"
             x:Name="MasterDetailsViewControl"
             ItemsSource="{x:Bind ViewModel.Items}"
-            SelectedItem="{x:Bind ViewModel.ActiveItem}"
+            SelectedItem="{x:Bind ViewModel.ActiveItem, Mode=OneWay}"
```
nigel-sampson: Thanks, for catching this, one time definitely wasn't right.  Although in my opinion this should be a two way binding, typically other code in the view model should be based off `ActiveItem` and that it should remain in sync with what the user has selected.

Issue #1175, opened 2017-09-21
Theme selection stops working when Frame navigates


mvegaca:
### Repro steps

Open [BugRepro App](https://github.com/Microsoft/WindowsTemplateStudio/files/1320689/BugRepro.zip)
Go to Settings Page.
Set Dark Theme
Go to Page1
Click on Navigate Button

### Expected Behavior

Navigate to Page 2 and Page 2 is on dark theme.

### Actual Behavior

Navigate to Page 2 and Page 2 is on light theme.

mrlacey:
> Open BugRepro App

That's not an adequate repro step. What is that app? How was it created?
It will need to be recreated for other frameworks & language to test that the issue exists and if necessary is corrected there too.

mvegaca:
I've create this app using tab project with MVVM Basic and C#. Added two blank pages and settings page. Blank 2 removed from pivot. Button added in Blank 1 to navigate from Blank 1 to Blank 2.

ralarcon:
Agree with @mrlacey. The repro steps must define all the steps required to reproduce the issue from scrach.

mvegaca:
Is like if the second navigated frame only applies selected framework to title TextBlock but the page background always take the RequestedTheme (App.xaml) color.

ralarcon:
Moving to 1.5 as it is not urgent.

mrlacey:
Moving back to m1.4 Bugs in the generated code should be considered high priority.

ralarcon:
Yes, you are right. We postponed it because of it is not a bug directly in the generated code, you need to make some changes / modifications to get a direct repro, in this case, you need to navigate out of the main frame to be able to get a repro. Anyway, we can keep it and if we are able, we'll do.

mrlacey:
While the bug doesn't occur when using the generated code it occurs because of it in a perfectly valid use case. The NavigationService code is causing the bug because it's being used in the way it was supposed to.

mvegaca:
Fixed on PR #1255

The problem was that we are not setting the ThemeResource background to pages and when the color changes and navigate the background does not apply theme background

sibille:
Verified in dev-nightly:
Wizard: 0.14.17284.65001
Templates: 0.14.17284.2

PR #1255, opened 2017-10-09
Issue1252 backgroundcolor

mvegaca:
#1175

Add style and background to all pages and controls

### DIFF
--- a/templates/Pages/MasterDetail.CaliburnMicro/Views/MasterDetailViewDetail/DetailsView.xaml
+++ b/templates/Pages/MasterDetail.CaliburnMicro/Views/MasterDetailViewDetail/DetailsView.xaml
@@ -4,6 +4,7 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    Style="{StaticResource ControlStyle}"
     mc:Ignorable="d"
     d:DesignHeight="300"
     d:DesignWidth="400">
--- a/templates/Pages/MasterDetail.CaliburnMicro/Views/MasterDetailViewDetail/MasterView.xaml
+++ b/templates/Pages/MasterDetail.CaliburnMicro/Views/MasterDetailViewDetail/MasterView.xaml
@@ -4,6 +4,7 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    Style="{StaticResource ControlStyle}"
     mc:Ignorable="d"
     d:DesignHeight="300"
     d:DesignWidth="400">
--- a/templates/_composition/_shared/Page.AddPageStyle/.template.config/template.json
+++ b/templates/_composition/_shared/Page.AddPageStyle/.template.config/template.json
@@ -1,0 +1,28 @@
+﻿{
+  "author": "Microsoft Community Contribution",
+  "classifications": [
+    "Universal"
+  ],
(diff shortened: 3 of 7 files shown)


mrlacey:
What's the point if this change?
What's it trying to solve?

This seems like massive overkill for setting the default background color on a page. 
Why not just set the background property on each page directly?
What are the consequences for creating a new style that is applied to each page? Doesn't this have an impact on how people will be able to extend the generated project if they want to change the styling of a page?

mvegaca:
I've added the BackgroundColor in a stylesheet because we are setting properties using styles instead of adding hardcoded values.

New pages added to new projects or on existing projects will use this style or use it if necessary

mrlacey:
I have big concerns about the way this has been done.

This has been applied to every regardless of the project type and yet it is trying to address something caused as a consequence of changes made specifically to address a question (#775) relating to SplitView projects. This seems unnecessarily broad. Inclusion in blank projects seems especially unnecessary as for that project type we hand all navigation responsibility over to the developer. It, therefore, seems wrong to be adding in code that fixes navigation issues in certain scenarios.

As a consumer of the generated projects, I have other concerns and questions

- Why does every **page** reference a style called **ControlStyle**? Is this something for styles or pages?
- Some pages also include other style definitions. As a developer working with the generated code, the inconsistency means it's not obvious where I am expected to add my own styles. 
- Some pages only contain a Grid with a BackgroundColor set. In these instances, setting the Page Background has no visible impact.
- In the Control.xaml file it sets the Page background with an explicit style. Could the same effect not be achieved with an implicit style on the Page?
- It's not clear why Control.xaml exists and what it's purpose is. This is especially unclear as removing it has no obvious effect. (Addressing this by adding a comment should be the last option considered.)
- As the control style only sets the page background, wouldn't it be simpler to set the value to the theme resource on each page?
  

The code that WTS generates is supposed to be of a high quality and demonstrate recommended/"best" practices. This does not do this.

ralarcon:
@mrlacey We opted to use a style resource file to avoid the proliferation of that style among the application pages and ease the change of the background if needed. You are right regarding the name, it could be more apropiated to use a file called "PageStyles.xaml" to have the page style. Then the controls can use a transparent color to allow the background page to set the background properly.

Regarding background compatibility, when you create a new page, the style page will be created if do not exists. Existing pages will not be affected (it is true that will not have the style applied, as they where created).

mrlacey:
This still doesn't address the why it's needed question

mvegaca:
It's needed because if this theme brush use the selected background color in the app, working on theme bug I've created a PoC from a UWP empty template and I've seen that changing the selected theme you can not see the controls if you remove this background brush. I don't know the moment in we stop using this background because it 's already added when a new app is created using the default empty project.

mrlacey:
My point is that it's not clear to anyone seeing it in the generated code.

ralarcon:
Maybe I miss something... It is in the style page and, as far as I understand, the usage of a style page is a best practice. If it is too obscure and hidden we can add it to each page... find and replace works fine too to make bold changes... but our understanding was that the style page was better.

mrlacey:
If you've no knowledge of this issue and you generate an app that includes this code it seems odd for the page background to be defined this way. There's nothing in the code to explain why this is there. (Note. again I'm not saying this needs a comment.)

Also, is this not possible with implicit styles?

ralarcon:
Let's do whatever makes more sense, is more understandable and at the same time address the issue. @mvegaca what do you think? can we address the issue using a more clear / understandable approach?

mvegaca:
Add the background in the page it's not a bad option, this is the default behavior on a new project creation.
image

The problem of not using a common page style is that if you want to set a new background for your app it may be changed in multiple pages. We are also trying to show how to use style resource dictionaries in the app. I think that is a good practice to add visual properties for all controls on styles (background, fonts, sizes,...) and add hardcoded values for properties related with the position of other elements in page (Margin, HorizontalAligmen, ...)

What do you think @ralarcon ?

mvegaca:
Updated on #1272

Issue #772, opened 2017-07-12
LiveTile Feature needs improved default activation handling code


mrlacey:
The default Live Tile code includes just comments in the `HandleInternalAsync` method in `ToastNotificationsService`. This is potentially confusing if you use the toast to launch the app and haven't modified this method. (It confused me.) Because it doesn't navigate anywhere once the app is launched it looks broken.

![image](https://user-images.githubusercontent.com/189547/28114749-0b81a3da-66fa-11e7-8377-ec323c7a08ce.png)

I think we should extend the sample code to include calling the NavigationService to go to the home page and include comments about querying the event args to determine where to navigate to and possibly passing params too.
_I'm aware that trying to go to the "home" page from right-click add of this feature might be an issue though._

mvegaca:
I don't think that feature template must be modified, I think that we could do a sample that uses LiveTile and notifications and implements custom activations.

crutkas:
@mvegaca lets discuss before instantly jumping to action.

mvegaca:
One possibility could be to use [TasksApp Sample](https://github.com/Microsoft/WindowsTemplateStudio/tree/dev/samples/suspendandresume/TasksApp) adding LiveTile feature to show different tasks. When the livetile is clicked open the app making the task on the livetile as active.
On this case, TasksApp sample would be a sample of suspend and resume and live tile features

mrlacey:
My issue here is that the code we generate by default seems, to me, to be incomplete. It doesn't create a working app.
Having been through lots of confusion with the way the ActivationService works while implementing #506 (protocol launch) I know there are a lot of potential issues with implementing this.

To say that this isn't a problem feels like you're saying that no-one else will have a problem with this. I may not be the smartest person in the world but I'd be amazed if everyone else who ever tries to use this feature has no problems, questions, or confusion that we can't help with by extending what's in the generated code.

To just have a sample that shows a complete version of this feels like we'd be saying "this code isn't complete but you can see a finished version at XXX." This feels very short of the mark for what we should be outputting.

ralarcon:
I agree with @mrlacey but at the same time I think that any code added to the Live Tile service to showcase the navigation will be needed to be removed as it will be sample code (the way to navigate or show a pop-up using the args). 

Complementing the current Live Tile feature with some documentation referenced in the "TODOs" will help users for sure. 

[...]

Issue #886, opened 2017-07-25
Add More Documentation around MVVM Basic Pattern


michael-hawker:
* **Overview:** It'd be great if there was a bit more structure to the MVVM Basic Getting Started section besides a link to Wikipedia.
* **Links:** https://github.com/Microsoft/WindowsTemplateStudio/blob/master/docs/getting-started-endusers.md

Ideally, the template/documentation should fill-in some back-end model objects and show how those hook to the existing View/ViewModel patterns in WTS.

There should also be additional info about how you have an over-arching application/command model so that things can be interconnected.

mrlacey:
[getting-started-endusers.md](https://github.com/Microsoft/WindowsTemplateStudio/blob/dev/docs/getting-started-endusers.md) also needs other updates:
- The pages sections need descriptions (could be copied from the templates.)
- Need to add entries for the pages and features added since this was last updated.

ralarcon:
Taking care of documentation update for Pages / Features.

We saw the MVVMBasic as a simplification of any existing frameworks so we understood the users were familiar with the pattern and its applicability to UWP apps.

@crutkas we need to plan to create the MVVMBasic documentation as a framework if we consider is worthy.

mrlacey:
I don't think we need to create a lot of documentation for MVVMBasic but needs something for people who need more information about how it compares to other options.

Just a few paragraphs should be enough. 
Something like:

MVVM Basic is not a framework but provides the minimum functionality to create an app using the Model-View-ViewModel (MVVM) pattern.
Use it if you can't or don't want to use a 3rd party MVVM Framework.

MVVM Basic is not intended to be a fully features MVVM Framework and does not include some features that other frameworks do. ViewModel-first navigation, IOC, and messaging being the most obvious ones. If you want these features then choose a framework that supports them.

Projects created with MVVM Basic contain two important classes, `Observable` and `RelayCommand`.
**Observable** contains an implementation of the `INotifyPropertyChanged` interface and is used as a base class for all view models. This makes it easy to update bound properties on the View.
**RelayCommand** contains an implementation of the `ICommand` interface to make it easy to have the View call commands on the ViewModel, rather than handle UI events directly.

ralarcon:
This could be part of the "more info" in the wizard and can be included in the getting started. @michael-hawker does this make sense?

mrlacey:
[...]

Issue #358, opened 2017-05-17
Feature: Share source


crutkas:
With the new Share contract, it would be great if this would be an option in the wizard.

https://blogs.windows.com/buildingapps/2017/04/06/new-share-experience-windows-10-creators-update/

mrlacey:
Should being a share provider and sharing content be separate features?
I expect more apps to want to share content than having content shared with them.

jamesmcroft:
@mrlacey I'd say yes to this for exactly the reason you've outlined.

mrlacey:
Making this feature about being the share source and a separate issue added for being the share target.
I expect these to be implemented together but they could be done separately. (This one probably being more useful and simpler to implement and test.)

mvegaca:
My idea here is to add one feature that adds some classes with the necessary code to share content and always adds one page with some contents (Texts, links, image, etc) and the possibility to share this content in different ways (Text, Image, HTML, Link, etc).
Here is a screenshot of my ShareSource PoC.

![image](https://user-images.githubusercontent.com/5556784/31658647-e400ebde-b331-11e7-803b-5c11fd2d5562.png)


It looks good fourn you @crutkas ?

crutkas:
Let’s set up time to better let me understand this.

mrlacey:
I'm not sure this feature makes sense.
All it would add would be sample code that demonstrates how to call an API.
Documentation for this already exists.
Anything/everything that's added will be deleted in the final app. It would be a dangerous precedent to start adding features that are just example code that shows how to use an API and not something that will be directly part of the final app.

mvegaca:
I agree with @mrlace this page could be removable code but demonstrate how to use an API is exactly what are we doing on LiveTile, ToastNotifications, etc.

mvegaca:
I'am thinking that share shouce could be a part of other page or feature in with it can all more valuew of this page or feature, for example, ImageGallery could have on the detail page a Command Bar proposed in #1099 with a Share button that allows to share this image with other apps in Windows 10.

mrlacey:
I've been thinking about this a bit more.
There is value in showing example code but we should do this in a way that makes it very clear that it is just an example. It shows how to do something but is not expected to be part of the final app. The other benefit is having something working within an app and within that app's structure and using the desired framework rather than a standalone example.
Any such example should probably be in a separate area within the wizard to make it clear that it is different. The frameworks list already has a level of subdivision and we could do something similar for pages.
[...]

PR #1268, opened 2017-10-10
Issue772 doc

mvegaca:
Issue #772
Added LiveTile doc and samples.
The samples include:
- Activation for LiveTile icon update.
- Activation for SecondaryTile icon.
- Different pages for SecondarySection and LiveTileUpdatePage.

The samples are implemented on CodeBehind, MVVMBasic and MVVMLight framworks.


mvegaca:
Could you please add a grammar review @mrlacey ??

ralarcon:
@mrlacey could you please add your review?

mrlacey:
I can't approve this.

1. It doesn't address the original issue.
2. It contains lots that seems redundant.
3. There is text in the documentation where I can't correct the grammar as I can't work out what it's trying to say.
4. It doesn't include a complete set of samples.

In reverse order.

We now have four officially supported frameworks. If we're providing framework specific examples this should include all the frameworks. It's not appropriate to require Nigel to write the CM version of everything. As we add support for more frameworks this will require learning those frameworks and/or collaborating on changes with people who are more familiar with those frameworks before submitting a PR to /dev.
Samples are also something we should avoid unless there's no other alternative. They're awkward and costly to maintain plus they are hard to point to from the generated code. We now have the situation where the generated code says "you need to do something, go look at this doc" (or it should--but I can't see it in the PR) then the doc says "now go look at this sample". This re-direction seems unnecessary and subject to more maintenance overhead over time.

The structure of the documentation can be summed up as
- The files that are added to the project
- Some of the code from the added files
- The code you must add to your project
- A link to samples of the above

This raises a number of questions in my mind
- Why include the copy of the code from the generated project? If it's not clear then it should be improved or comment in the code added.
- If the code that needs to be added is in the document, why is the sample needed?
- If the code must be added in every app why can't it be included in the generated code?.....


And now the fundamental issue.
I raised #722 as it feels like the feature only contains part of what's necessary for that feature.
The generated project that includes that feature isn't complete and doesn't work fully.

The Live Tile feature should (IMHO) include three things
1. Code to pin or update a tile
2. An example of how to call the code added in 1
3. Handling activation of the app from a tile.

We currently only do 1 & 2. The original issue is that we should include 3 as well.

Every app that uses the feature will _have_ to implement the activation code. Not including a base version of the activation handling code but requiring everyone to have to go and look at the docs and copy from there does not feel like the best solution we can provide.
As with most of the other code we generate, we should be providing something that works but with TODO comments in the code to show where changes have to be made.

When I did the UriScheme feature I deliberately included the activation handling because
- it meant there was less for the developer to do
- it wasn't necessary to require them to go and look elsewhere to know what needed to change
- it was easier to test what was included in the generated test.

The last point was particularly important for me as I needed to manually verify that the generated code worked correctly. If I hadn't included that code I've needed to manually add it for each variation of output (projectType/framework) to do the testing. (Did anyone manually add that code to test the feature when it was originally added? I'm skeptical and concerned that if we don't do this it will make adding automated testing of the generated feature harder.)

There are other features that also suffer from a lack of including the logic to handle activation. I think we'd benefit from reviewing/discussing/improving these all together. There _may_ be a benefit for some reuse in the templates needed to add all the missing activation handlers but this may be missed if they're reviewed separately.

@crutkas @ralarcon we might want to turn this into a bigger discussion / separate issue. Keen to talk more about this.

ralarcon:
I think the PR try to cover what whas defined in the #772, probably not in the best way, but at the end of the thread @mvegaca clearly expressed that he was going to add some documentation to help understand the usage. No one expresed against that, so we went ahead.

@crutkas, @mrlacey close this PR if you think does not have value. If not, please, let me know which part of the content are understandable and we try to reauthor.

I agree with the fact of adding the pages / features with sample code if required is more maintainable than create samples, but this goes directly against the original concept of avoid to create non-final code and avoid the developers to remove generated code. We can do it, we just need to define it as the default. In the other hand, our audience is the developers, I think the samples are perfectly understandable without having to include variations for all frameworks and project types.

mrlacey:
This whole discussion raises the need to be clearer about what the goals of the project are and how to achieve them.

There's lots of contradiction going on here and it's not productive.

> "avoid to create non-final code and avoid the developers to remove generated code"

This contradicts the fact that we already include "sample" code for LiveTile, Toast Notifications, and anything using SampleData. We do it because what would be generated otherwise would be half complete.
There are also lots of places where we add code with a TODO comment saying it needs changing or implementing. This is "non-final" but is both necessary and beneficial.

My argument is that this is half complete as it doesn't include details of how to handle the activation.
I acknowledge that we can't implement everything as it will be needed in the final app but that doesn't mean we should do nothing.
All I was asking for in the original issue was some code in the `HandleActivationAsync` method in `LiveTileService` (and `ToastNotificationService`) so that launching the app wouldn't leave it in a state where no page was navigated to. In this scenario, the app is in a state which it should never be and can easily be considered a bug. As another of the goals of the project is to output a working app. While it's not a final app it should work without behavior that looks like a bug.

Being clear about the aims and what we can and should include would also benefit #358 which has turned into the gathering of ideas for example code to generate. That is code that is intended only to serve as an example of how to use a specific API. That is code that should also be deleted and so is non-final.


> In the other hand, our audience is the developers, I think the samples are perfectly understandable

It's always dangerous to make assumptions about the people who will use a piece of software as they will vary widely in terms of knowledge and experience. We know this is especially true as WTS is targeted at those just getting started as well as experienced developers. We also know from issues like #886 that people want help with things that we might consider basic.
Not all the developers who use the tool will think about or understand code the same way as you.
If we're going to have samples then not covering the different ways that the code should behave with different frameworks will mean that some people will be unlikely to follow the sample. But, I still feel it's better to not need samples at all.

I also don't agree that not objecting to a subjective statement about one way to address an issue is the same as approving that it should be done that way.  I don't have the time or desire to review and respond to every comment on every issue. I already spend more time reviewing changes and responding to issues than I do working on improvements directly and that's not how I want to spend the time I'm volunteering.

ralarcon:
@mrlacey I hear you. I'm the first one interested in delivering quality and value to our developers.

Probably I didn't select the most appropiated wording. Instead of "avoid", I should have used "reduce". That is, to reduce as far as possible the use of non-final code. This is what is in our minds, but of course, the measurment of what is the amount of acceptable sample code depends on the eyes, background and experience of who is looking at the code.

Thinking in positive, I don't have any interest in supporting documentation as it is more difficult and costly than having working code. The code for the sample was there just to support the doc. 

My assumptions arround developers were in the sense that even having different framworks, my understanding is that it is not absolutely mandatory to have al the variations for a sample as the samples inplace were not too complex. This happen even in docs.microsoft.com, not all [samples](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.datatransfer.sharetarget.shareoperation) are always in all languages. 

Finally, don't get me wrong, we appreciate and value your criticism as it make us to be better. I wasn't saying that the way to afford the issue was approved and accepted by all... As you mention the issue (and that the PR didn't address it), I was just mentioning that the PR was in line with what was discussed in the issue, even if it wasn't the best / ideal approach, nothing else. 

This is not dramatic for us to close this PR, if we think we can achieve the target using other way.

Issue #1232, opened 2017-10-04
Missing ImageGallery page icon and descriptions


mvegaca:
Missing ImageGallery page icon and descriptions

ralarcon:
@crutkas are you going to create the icon and the descriptions? If not, let me know and we can take care.

crutkas:
I want this page reviewed by design team and a few others before we fully ship it.  I don't think we have time for 1.4 to be inclusive here

ralarcon:
Ok, by now, we will include the image gallery for "final" testing purposes but we will not ship it (making it unavailable). You can request the review. We finally ready, we can update the templates to include again and ship an update.

mvegaca:
Make as hidden on #1274

dgomezc:
@crutkas for the gallery page icon we can use this:

![image](https://user-images.githubusercontent.com/12777039/31765406-8be6916c-b4c4-11e7-9959-52dbe0161bd2.png)

It has been created by http://modernuiicons.com and is licensed CC. The camera icon was created from this source as well.

crutkas:
can we make this smaller?  very dominate

crutkas:
we also need to attribute back to them.

ralarcon:
Released in 1.5

PR #1274, opened 2017-10-10
Make image gallery as hidden

mvegaca:
#1232


ralarcon:
I'm thinking that if the design team wants to review the image gallery, they (or we) must be able to generate an app using the image gallery... this change could be done in master... so this PR should be against master once we merge the contents we currently have in dev.

ralarcon:
Changing merge target to master as we want the Image Gallery page disabled only for pre-release and release versions.

PR #1332, opened 2017-10-23
Issue365 sharetarget

mvegaca:
Add MVVMLight, CodeBehind and CaliburnMicro frameworks to ShareTarget feature.

@nigel-sampson can you review the feature on CaliburnMicro.
I've found differences in App.xaml.cs files and I've added a different template ShareTarget.CaliburnMicro. I was having errors on ViewModels registration when App was launched from ShareTarget activation and I fixed it with a workaround on ExamplePage ViewModel creation.


nigel-sampson:
Will review the Caliburn.Micro parts of this tonight.

templates/Features/ShareTarget.CaliburnMicro/App_postaction.xaml.cs
```diff
@@ -0,0 +1,33 @@
...
+//{[{
+
+        protected override void OnShareTargetActivated(ShareTargetActivatedEventArgs args)
+        {
+            // See more about configure share target in UWP
+            // https://docs.microsoft.com/en-us/windows/uwp/app-to-app/receive-data
+            // See also how to share data from you App
+            // https://docs.microsoft.com/en-us/windows/uwp/app-to-app/share-data
+
+            var rootFrame = Window.Current.Content as Frame;
```
nigel-sampson: This block side steps activation / creation of navigation services leaving the app in a mostly uninitialized state. I'd recommend going through the activation handlers.

templates/_composition/CaliburnMicro/Feature.ShareTargetExamplePage/Views/wts.ItemNameExamplePage.xaml.cs
```diff
@@ -0,0 +1,27 @@
...
+        public wts.ItemNameExamplePage()
+        {
+            DataContext = ViewModel;
+            InitializeComponent();
+        }
+
+        protected override async void OnNavigatedTo(NavigationEventArgs e)
+        {
+            base.OnNavigatedTo(e);
+            await ViewModel.LoadAsync(e.Parameter as ShareOperation);
```
nigel-sampson: For Caliburn.Micro if you end up passing the `ShareOperation` as parameter into the navigation service (through activation handler) the nav service can directly populate a property `Parameter` on the view model without requiring any code behind.

templates/_composition/CaliburnMicro/Feature.ShareTargetExamplePage/ViewModels/wts.ItemNameExampleViewModel.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            get => _sharedData;
+            set => Set(ref _sharedData, value);
+        }
+
+        public wts.ItemNameExampleViewModel()
+        {
+        }
+
+        public async Task LoadAsync(ShareOperation shareOperation)
```
nigel-sampson: You can use the `OnInitialize` method on the view model to call this to wire it into the view model life cycle.

templates/_composition/CaliburnMicro/Feature.ShareTargetExamplePage/Views/wts.ItemNameExamplePage.xaml
```diff
@@ -0,0 +1,62 @@
...
+        <Grid Margin="{StaticResource MediumLeftRightMargin}">
+            <TextBlock
+                x:Name="TitlePage"
+                Text="{x:Bind ViewModel.PageTitle}"
+                Style="{StaticResource PageTitleStyle}" />
+        </Grid>
+
+        <ContentControl Grid.Row="1" Content="{x:Bind ViewModel.SharedData, Mode=OneWay}">
+            <ContentControl.ContentTemplateSelector>
+                <templateSelectors:SharedContentTemplateSelector>
```
nigel-sampson: I tend to stay away from template selectors in Caliburn.Micro as the framework has better support built in. Take a look how the tabbed pivot locates different views for the view models. 

If you don't want to create different view models per share type (which IMO makes sense rather than a hold all `SharedContentModel`) then you can use the `DataFormat` as a context for view location much like master details does.

mvegaca: I think I don't understand this. The idea is to have the same ViewModel and show different layout (in the same Page) depending on the kind of data that is shared with the application.

nigel-sampson: This might be a subjective design decision, but a single view model for all share types feels a bit off. normally I'd recommend a `SharedViewModelBase` that holds common properties such as `Title` and other metadata with a `SharedWebLinkViewModel` inheriting from that.

This way the view only needs to contain the following xaml. CM will then find the appropriate view (in the above case `SharedWebLinkView`) and insert it  This nicely separates the display of the different types share data from each other. It also means we won't need the template selector.

``` xml
<ContentControl cm:View.Model="{Binding SharedData}" />
```

If you want to stick with a single view model for all share data then I'd recommend something like the following in the xaml.

``` xml
<ContentControl cm:View.Model="{Binding SharedData}" cm:View.Context="{Binding SharedData.DataFormat}" />
```

This affects the view location with a context on the format so if the data format was WebLink then the inserted view would be something like `/Views/SharedData/WebLink.xaml`. Again separating the views and discarding the template selector.

Issue #863, opened 2017-07-21
Port existing C# templates to VB


mrlacey:
As a continuation of #371 and X-Ref https://github.com/Microsoft/WindowsTemplateStudio/wiki/Visual-Basic

## Port existing C# templates to VB.

**Pages**
- [x] Blank
- [x] Settings
- [x] WebView
- [x] Media Player
- [x] Master/Detail
- [x] Grid
- [x] Chart
- [x] Tabbed
- [x] Map
- [ ] Image Gallery ?
- [ ] Camera

**Features**
- [x] Settings Storage
- [x] Suspend and Resume
- [x] BackGround Task
- [x] Toast Notifications
- [x] Azure Hub Notifications
- [x] Dev Center Notifications
- [x] Live Tile
- [x] First Run Prompt
- [x] What's New Prompt
- [x] URI Scheme
- [ ] Launch Feedback Hub ?
- [ ] Share Source ?
- [ ] Share Target ?
- [ ] DragDrop Source ?
- [ ] DragDrop Target ?  

## Need strategy and automation for the handling of future changes to C# templates and localization.

**Needed (PowerShell?) scripts**
- [x] Identify missing VB pages/features
- [x] Identify changes to CS pages/features/projects (based on file last modified time)
- [x] Identify changes to CS pages/features/projects (based on git commits)
- [x] Process to copy description changes and translations from CS equivalent config files
  

**Needed automated tests**

Some of these may need to be purely manual so that C# code can evolve ahead of VB...TBD

- [x] Check non-code content is identical in projects created with the same options
- [x] Check both using the same versions of 3rd party libraries
- [x] Check comments are the same in equivalent files
- [x] Structure, formatting, coding conventions, etc. of generated VB code (via SonarLint)
- [x] Check equivalent code files contain equivalent elements (via Roslyn)
- [ ] Check equivalent generated projects function the same way (via WAD)
- [ ] Check equivalent generated project create the same visual output (via WAD)
  
  
  

## Need to update documentation for VB support.

**Update**
- [ ] Main readme
- [ ] CONTRIBUTING
- [ ] Activation
- [ ] GettingStarted *n
- [ ] Templates
- [ ] docs/projectTypes/navigationpane.md
- [ ] docs/features/whats-new-prompt.md
- [ ] docs/pages/settings.md *n
- [ ] anything else with code samples?

mvegaca:
Hi @mrlacey @ralarcon @crutkas 
[...]

PR #1334, opened 2017-10-23
Add VB Templates

mrlacey:
For #863 - which will be split into multiple other issues for outstanding work.

This adds the file new option in VS and the generation of everything included in 1.4 apart from the Camera Page.

There's a potentially large impact on testing with this as most tests are now run for both languages. This, inevitably, means they take longer to run.
There are also some new VB specific tests and some ManualOnly ones intended to ensure that the generated code functions the same way, regardless of language.

PR raised across multiple forks because of an issue with the original fork being linked to a PR that doesn't exist


mrlacey:
These failing tests run fine on two other machines I have. 
Tests are all the ones that relate to generating a VB project. Investigating if this is an appveyor config issue.

sibille:
@mrlacey, I'm not sure if this is the cause, but might be related. 

I'm looking at the FakeSolution to see how to allow various projects being added to the solution and just saw that ProjectTemplateVB is not set correctly, VB.NET project type is F184B08F-C81C-45F6-A57F-5ABD9991F28F, according to docs (https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/solution-dot-sln-file)

ralarcon:
Trying to adjust the appveyor build to output the generated (failed) projects as a build artifact to be able to analyze what has been generated. Running the build for this PR again... let see if now the output is available.

mrlacey:
@sibille well spotted. I don't know where the other one came from
@ralarcon I've duplicated the AppVeyor Build and am also investigating. Running with detailed output killed Chrome though

mrlacey:
Think I've identified the issue..testing now

ralarcon:
Looks like something related to AppVeyor. The artifacts in the output shows like the VB projects not being generated properly. I've tested this branch in my computer and the tests runs properly...

mrlacey:
The first time I ran from the command line on my machine I was able to recreate the failure and noticed that the console runner was on an older version number than the VS runner. making these the same made the local issue go away. Obviously, this didn't address the issue on the server and switching back to the old command line runner on my machine no longer causes the issue. This was a red herring and not the issue.
Investigations continue

sibille:
> Microsoft.Templates.UI.GenException : Template 'Default', name: 'TabbedPivotCodeBehindAllRandomVB', reason: 'Error while processing file /Projects/DefaultVB/wts.DefaultProject.vbproj
> Non-negative number required.
> Parameter name: srcOffset'

This error is thrown on generation of the vbproj file in the template-engine https://github.com/dotnet/templating/blob/master/src/Microsoft.TemplateEngine.Core/Util/Orchestrator.cs 

We have verified that the same code (https://github.com/Microsoft/WindowsTemplateStudio/tree/Issue863-VBTemplates) works in our VSO Build Servers, that confirms that it seems to be sth. appveyor-related. 
Maybe we should set up a separate execution set for vb tests, and not execute them in appveyor until we find the root cause.

mrlacey:
@sibille @ralarcon As I have been unable to identify why this continues to fail I think it's worth considering disabling the VB tests as part of the CI process.
Assuming this is a temporary solution due to either an AppVeyor config issue or a combination of AppVeyor plus an old version of the template engine it would be good to be able to enable it at some point in the future.
Temporarily disabling the VB tests on just the CI server will be harder than just separating them entirely but I'm wondering if you'd prefer that anyway as it will make things easier for developers just working on the C# templates.
So, 

Option 1. Would you like the VB tests disabled for the CI server in a way that is easier to re-enable them at some point in the future?

OR

Option 2. Would you like the tests that run on the VB templates to be more clearly separated from the ones that run on the C# versions? (This will mean they won't get picked up by the existing CI config by giving them new Traits.)



Option 1 will be harder as I can't see a simple way to do this. Option 2 will make the code simpler but will lead to more of it and is likely to take more effort to integrate into the CI process if and when you later want to do this.

mrlacey:
Making the following changes to separate the VB project generation tests from those that are part of the gated set to avoid the current issue and require minimum changes.

- ExecutionSet 'BuildMinimum' no longer includes the VB version of those tests. New ExecutionSet 'BuildMinimumVB' includes the VB versions.

- ExecutionSet 'Minimum' and all other 'Build*' ExecutionSets will continue to generate both language versions.

- SonarLint tests are in new ExecutionSet 'BuildVBStyle' and no longer in ExecutionSet 'BuildStyleCop' rather than 'BuildSonarLint' as I'm planning to have more than just SonarLint Analyzers for VB code style checking.

Issue #365, opened 2017-05-18
Feature: Share target


mrlacey:
Add a feature which includes the functionality to enable the app to be the target of sharing.

Split from #358

crutkas:
@mrlacey bump both of this to 1.4, you ok with that?

mrlacey:
bumping is good with me

mvegaca:
PR added #1324 

MVVM Basic only, I want you can check it out in Dev Nightly to can give opnion and improvements before translate ir to other frameworks.

Feature allows to share Images (Edge) and Photos (Windows Photos) with the generated app with Share Target declaration.

Implementation includes:
- Activation from share target in App.xaml.cs
- ShareData extension file
- Model data
- DataTemplateSelector to show different layout between Web Link and Images share layout.

mvegaca:
@crutkas @mrlacey have you look that feature on MVVM Basic? Do you have any comments or improvements?

mrlacey:
haven't had a chance to look at it yet

mvegaca:
Translated to MVVMLight, CodeBehind and CaliburnMicro in #1332

mvegaca:
@nigel-sampson can you review the feature on CaliburnMicro??
I've found differences in App.xaml.cs files and I've added a different template ShareTarget.CaliburnMicro. I was having errors on ViewModels registration when App was launched from ShareTarget activation and I fixed it with a workaround on ExamplePage ViewModel creation.

ralarcon:
You are right mrlacey, I personally asked to finish the template (even re-work needed) as it was no so much effort. We didnt't assume tacit approval.

A the same time, I have decided that, before creating a template, and in order to have better discussions arround features, we will share the approach using a PoC app.

mrlacey:
So, having a look at the output in DevNightly (0.14.17296.1)
_Because I've been forced into reviewing this now before it's marked as complete without anyone reviewing it._

- No icon
- No short description (it's lorem ipsum)
- No detailed description (this may be a good place to link to more information, etc.)
- As per #827 this feature should not be listed under "User Interaction"
- ShareTargetExamplePage includes unnecessary using statements
- SharedContentTemplateSelector is used only by ShareTargetExamplePage but there's nothing to 
- ShareTargetExamplePage includes localized resources that will be left behind once the example page is deleted.
[...]

PR #1347, opened 2017-10-24
Updates on Share Target Feature

mvegaca:
#365
Updates on Share Target Feature

### DIFF
--- a/templates/_composition/_shared/Feature.ShareTarget/Activation/ShareTargetActivationHandler.cs
+++ b/templates/_composition/_shared/Feature.ShareTarget/Activation/ShareTargetActivationHandler.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using System.Threading.Tasks;
+using Windows.ApplicationModel.Activation;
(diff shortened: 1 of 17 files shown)


templates/_composition/_shared/Feature.ShareTarget/Activation/ShareTargetActivationHandler.cs
```diff
@@ -0,0 +1,21 @@
...
+using Windows.UI.Xaml.Controls;
+using Param_ItemNamespace.Views;
+
+namespace Param_ItemNamespace.Activation
+{
+    internal class ShareTargetActivationHandler : ActivationHandler<ShareTargetActivatedEventArgs>
+    {
+        protected override async Task HandleInternalAsync(ShareTargetActivatedEventArgs args)
+        {
+            var frame = new Frame();
```
nigel-sampson: This will have issues in Caliburn.Micro as you're wiping out the `Frame` being wrapped by the navigation service which has a lot of flow on effects around things like lifecycle.

Issue #1360, opened 2017-10-29
VBProj not getting updated as part of right-click add


mrlacey:
When adding a page or feature, to a VB project, that includes additions to vbproj (e.g. adding NuGet reference) the modifications to the vbproj file are not made and so the project is left in a state that does not compile.

**Repro steps**
File>New
- Navigation Pane
- Code Behind
- NO extra pages or features

RightClick>Add Page
- Chart page

The generated project doesn't include reference to Telerik NuGet package even though it's on the generation summary.
Open ChartPage.xaml and it shows errors due to no reference
Even rebuilding the project doesn't fix it.

Also happens for features, as well as pages

**All generated output should compile and work as expected**

Need to also understand why related tests pass. Are they giving a false positive? If so, need to check if other tests are also giving false positives.

mrlacey:
I can't recreate for certain but I think the difference between the test and real use within VS comes down to the way that VS holds references to files (esp. the project file) and the way they're updated in the background while open.
VSEmulator behaves in the same way and adding features to an existing project does add the reference.
The test is doing the same thing as when used within VS but was producing a different result. I don't think this is a reason to change the test but may be a good argument for more manual testing within VS or some automated tests that control VS to test generation.

sibille:
Verified on dev-nightly: 
Wizard: 0.14.17307.65003
Templates: 0.14.17307.3

PR #1361, opened 2017-10-29
Fix VB Right-Click package reference bug

mrlacey:
For #1360

Ensure that package references are added to the project as part of a right-click>add action are correctly saved and the updated project will build correctly.

Also restored packages when adding anything new as part of a right-click>Add action so the designer works correctly and references are shown in the solution explorer.


mrlacey:
Not sure why this failed. The same branch builds and passes test on my two machines and on[ my AppVeyor comparison build](https://ci.appveyor.com/project/mrlacey/windowstemplatestudio/build/1.0.32)

ralarcon:
looks like transient error... re-queued the PR (as the line where was failling have a control over null value) and didn't fail... not sure way. If this gets reproduced we'll need to investigate further.

PR #1397, opened 2017-11-03
Create drag and drop example

dgomezc:


crutkas:
why is this getting dragged into the main dev branch?

ralarcon:
It is in the poc folder... no matter the branch itself. If you prefer we can use a branch instead of a folder.

Issue #1390, opened 2017-11-03
Make Uri Scheme available in Caliburn.Micro framework


ralarcon:
Port Uri Scheme feature to Caliburn framework.

mvegaca:
Working on this

mvegaca:
Added on #1399

mvegaca:
Checked on **Dev Nightly**
Wizard: 0.14.17314.65001
Templates: 0.14.17314.1

PR #1399, opened 2017-11-06
Issue1390 urischeme

mvegaca:
#1390


ralarcon:
@nigel-sampson can you have a look on this just to double check it is proper aligned with Caliburn.Micro?

Issue #1407, opened 2017-11-06
Missing dots in Package.appxmanifest when generating with dev-nightly


sibille:
### Repro steps
Generate project with dev-nightly

What did you do? How can someone else reproduce this?

### Expected Behavior
Metadata in package.appxmanifest should be:

 <genTemplate:Metadata>
    <genTemplate:Item Name="generator" Value="Windows Template Studio"/>
    <genTemplate:Item Name="wizardVersion" Version="0.14.17310.65001" />
    <genTemplate:Item Name="templatesVersion" Version="0.14.17310.1" />
    <genTemplate:Item Name="projectType" Value="SplitView" />
    <genTemplate:Item Name="framework" Value="CodeBehind" />
  </genTemplate:Metadata>

### Actual Behavior
Metadata in package.appxmanifest is:

 <genTemplate:Metadata>
    <genTemplate:Item Name="generator" Value="Windows Template Studio"/>
    <genTemplate:Item Name="wizardVersion" Version="141731065001" />
    <genTemplate:Item Name="templatesVersion" Version="14173101" />
    <genTemplate:Item Name="projectType" Value="SplitView" />
    <genTemplate:Item Name="framework" Value="CodeBehind" />
  </genTemplate:Metadata> 

## System
Happens only in vsix, not in VSEmulator.

Wizard: 0.14.17310.65001
Templates: 0.14.17310.1

mvegaca:
Checked on **Dev Nightly**
Wizard: 0.14.17314.65001
Templates: 0.14.17314.1

PR #1414, opened 2017-11-07
Add v to version number to avoid conversion to int

sibille:
Quick summary of changes
Add v to version number to avoid conversion to int
#1407


ralarcon:
This workarround is temporal in order to check if there is any side effect of adding the "v" before the version and until we have a clear understanding of the root cause. X-Ref: dotnet/templating#1322

PR #1443, opened 2017-11-14
Reviewing Prism integration

bartlannoeye:
## PR checklist

Quick summary of changes

General review of the Prism generated samples. For some changes, I based myself on refactoring the sample app first and then apply it in the templates. I haven't had time yet to rerun all tests and try to regenerate the apps, so it's best to double check all changes.

Changes applied:
- Using nameof instead of string literals
- Apply correct ordering for `override async`
- Async methods should end on -Async
- Adding some useful comments
- Apply consistent variable naming
- Change unity registrations for singleton objects

Points to verify:

- [ ] Generated solution file encoding, this did change on editing. Regular code files don't have issues. See https://github.com/Microsoft/WindowsTemplateStudio/commit/b32cf65aa6f8a4587afecc213e0fc6638bf7b362
- [ ] Added comments to where I think they do work in the generation. Is this enough or do they have to be copied to all templates?
- [ ] Noticed generated app has LiveTileService vs LiveTileFeatureService in template. Similar for several other features. Is this changed through the template engine? Double check if my changes to Unity registrations are correct and not impacted.
- [ ] In App.xaml.cs is the await Task.CompletedTask intentional for the tooling to work (having the same method definition)? Otherwise return Task.CompletedTask would be better.
- [ ] Prism supports multiple IoC containers. I think we can launch with Unity and look into the other containers later (or even wait for Prism v7 as this will simplify some of the effort).

But in general: great work from the team !

### DIFF
--- a/code/src/Core/Mvvm/RelayCommand.cs
+++ b/code/src/Core/Mvvm/RelayCommand.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Templates.Core.Mvvm
 
         public RelayCommand(Action execute, Func<bool> canExecute)
         {
-            _execute = execute ?? throw new ArgumentNullException("execute");
+            _execute = execute ?? throw new ArgumentNullException(nameof(execute));
             _canExecute = canExecute;
         }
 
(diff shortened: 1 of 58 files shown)


ralarcon:
Added @LeeParrishMSFT as reviewer. Lee, once reviewed on your side you can merge to PRISM feel free to make a PR to dev when ready.

PR #1458, opened 2017-11-20
Add ShareSource feature dependency to WebView and ImageGallery pages

mvegaca:
#358


ralarcon:
We are not going to generate the addtions to share source to the pages to avoid developers the need of removing code if don't want to share the contents of their pages.

PR #1537, opened 2017-12-11
Update activation shell creation Fix #1533

### LINKED ISSUE #1533: ShellPage is instantiated on BackgroundTask activation while app is terminated
### Repro steps

1. Create a new app using **Windows Template Studio**.
2. Choose **Navigation Pane** and add a **Background Task**.
3. Open project properties and check **Do not launch, but debug my code when it starts**  in **Debug tab**
4. Start Debugging the project.
5. Start the app outside the VS and close the app (it will register the background task).
6. Start the background task from **Debug Location Toolbar** (background task must start while the app is terminated)

### Expected Behavior

ShellPage shouldn't be instantiated when the app is started only for background operation.

### Actual Behavior

ShellPage is instantiated in ActivationService constructor. It causes additional overhead for background task. It can get event worse depending on code executed with ShellPage initialization. For example some UI related method can throw an exception (e.g. **DisplayInformation.GetForCurrentView()**) when they are called in background task.

## System

* **VS Version:** 2017
* **WTS Wizard Version:** 1.5.17334.7
* **WTS Template Version:** 1.5.17334.7
* **Windows Build:** Any

mvegaca:
Missing fix it on Prism


mvegaca:
This PR also fix the bug on Visual Basic and I've checked that it's working as expected. Could you please add a review on VB implementation @mrlacey ?

Issue #1494, opened 2017-11-27
Add support for UseNavigationViewWhenPossible


nmetulev:
## For new Features:

* **Overview:** With the 2.1 update of the UWP Community Toolkit, the HamburgerMenu now support NavigationView. However, the WTS templates will need to be updated to ensure the HamburgerMenu will look great when using (and not using) the NavigationView

* **Links:**
[HamburgerMenu NavigationView documentation](https://docs.microsoft.com/en-us/windows/uwpcommunitytoolkit/controls/HamburgerMenu#navview)
https://github.com/Microsoft/UWPCommunityToolkit/issues/1652

mvegaca:
Working on it.

mvegaca:
Hi @nmetulev 
I'm updating the tookit to 2.1.0 version. In case of Hamburguer menu I've seen that if it's on FCU W10 appears a header text when the window is in narrow state.

This is a screenshot of toolkit sample app.
![image](https://user-images.githubusercontent.com/5556784/33551915-4eeacb04-d8f3-11e7-9d41-b3a0d5c9f8d7.png)

How can I customize this text or avoid this behavior?

mvegaca:
Could it be the NavigationView header that it isn't set to a HamburguerMenu property?

nmetulev:
It's a known issue actually, here is the fix: https://github.com/Microsoft/UWPCommunityToolkit/pull/1655

ralarcon:
Do you know when is planned to release this fix?

nmetulev:
If you need this before February, we can do a 2.1.1 release to add this fix this month hopefully.

crutkas:
@nmetulev we would need this before else it is a blocking issue.

mvegaca:
I've been updating Toolkit version to 2.1.1 on [this issue branch](https://github.com/mvegaca/WindowsTemplateStudio/tree/Issue1494-navigationview) and adding conditional Xaml on window size change visual state changes.

I've found an app crash applying these changes to new Prism framework, It looks an unhandled exception when HamburgerControl sets the items source to NavigationView. I've been attached a crash repro [here](https://github.com/mvegaca/WindowsTemplateStudio/tree/Issue1494-navigationview/poc/HamburguerControl/PrismBugRepro).

@nmetulev can have a look on this? I think the we are adding the same code on different frameworks and it still crashing on Prism. Thank you much!

crutkas:
[...]

PR #1590, opened 2017-12-19
Update Toolkit to 2.1.1 Issue 1494

mvegaca:
Add Toolkit 2.1.1 using NavigationView in Navigation Pane project type (UseNavigationViewWhenPossible property).
Issue #1494

With this PR we are adding an execution crash on Prism Framework.


mvegaca:
Setted Prism project to 10.0.16299.0 version, with this it's working on Prism!

Issue #1548, opened 2017-12-13
Prism: Adapt MasterDetail to use UWPCommunityToolkit and new designs


sibille:
We changed the Master Detail templates to use the ~~HamburgerMenu~~ MasterDetailsView control from the UWP Community Toolkit and did design changes. 
I think the Prism version of Master Detail should follow the same approach.

crutkas:
what caused us to change the template to use the hamburger menu?  when was this change done?

sibille:
@crutkas, I was refering to the MasterDetailsView control from UWP Community Toolkit, sorry for the confusion. 
We did this change in PR https://github.com/Microsoft/WindowsTemplateStudio/pull/1222/ and discussion on this took place in issue https://github.com/Microsoft/WindowsTemplateStudio/issues/1184

crutkas:
i missed everyone being in agreement this was the best way forward.

crutkas:
this feels like a prism ship blocking issue if the templates aren't in sync

ralarcon:
@LeeParrishMSFT I think this is ready to be closed, isn't it?

LeeParrishMSFT:
@ralarcon  Yes, we can close this as the fixes are checked in and verified in the dev branch.

PR #1594, opened 2017-12-20
Updated templates to use Tookit support for Master Details.

LeeParrishMSFT:
Updates for #1548 modifying the MasterDetail template for the Prism framework to use UWP Toolkit MasterDetailsView control.


LeeParrishMSFT:
I found the issue, somehow missed a change to remove a file that wasn't needed any more. I've corrected this, run the unit tests locally, and updated the pull request.

mvegaca:
Hi @LeeParrishMSFT I've check out the PR branch and I've generated an app using navigation pane project, Prism and Master-Detail Page named "Companies".
It looks good but the master-detail page it is not showing the page name.

It looks that Resources.Resw postaction it's not applying the correct item name.

image

image

Can you have a look here?

LeeParrishMSFT:
I've added the missing resources so the page titles will show correctly.

Issue #1586, opened 2017-12-19
Prism apps crashes on BackgroundTask activation while app is terminated


mvegaca:
This bug only happens on Prism, it's already fixed on other frameworks.
xref #1533

### Repro steps
- Create a new app using Windows Template Studio and Prism Framework
- Choose Navigation Pane and add a Background Task.
- Open project properties and check Do not launch, but debug my code when it starts in Debug tab
- Start Debugging the project.
- Start the app outside the VS and close the app (it will register the background task).
- Start the background task from Debug Location Toolbar (background task must start while the app is terminated)

### Expected Behavior
ShellPage shouldn't be instantiated when the app is started only for background operation.

### Actual Behavior

ShellPage is instantiated in ActivationService constructor. It causes additional overhead for background task. It can get event worse depending on code executed with ShellPage initialization. For example some UI related method can throw an exception (e.g. DisplayInformation.GetForCurrentView()) when they are called in background task.

The fix on other frameworks was adding a Lazy initialization on shell page to delay the page creation to the moment when it's really needed. It does not work on Prism Framework.

Solution code is detailed on #1533

crutkas:
@LeeParrishMSFT

milazzom:
I have a fix that I'll be checking in tonight to work around this issue in the template.

milazzom:
@crutkas What is the process to close out an issue like this after it's been merged?  I closed it out earlier, and noticed another issue we addressed was tagged as "can close out soon".  Thanks!  I'm re-opening this issue to see what the correct next-step is.

crutkas:
Close it out if you can verify in the nightly.

ralarcon:
@milazzom We mark as "Can Close Out Soon" those issues we solve but need to be validated in a nighly build. When validated, we add the wizard and template version numbers where we have done the validation and close the issue.

I will mark this that way to make sure we validate it within a nightly.

ralarcon:
Verified in Nightly
Wizard: v0.15.18012.65002
Templates: v0.15.18012.5

PR #1606, opened 2017-12-20
Adds logic to init the Unity container.

milazzom:
This PR resolves a crash condition for issue #1586.  The original crash condition reported in the issue was not encountered.  This change adds logic to invoke the appropriate methods in the base class to initialize the Container variable if it is null when a background task is launched.


LeeParrishMSFT:
Another option for this is to use the CreateAndConfigurateContainer method on the base PrismApplication Class:

        protected override void OnBackgroundActivated(BackgroundActivatedEventArgs args)
        {
            // Work around for Container not being initialized when launched from background task activation
            base.CreateAndConfigureContainer();

            base.OnBackgroundActivated(args);
            Container.Resolve<IBackgroundTaskService>().Start(args.TaskInstance);
        }

By calling base.CreateAndConfigureContainer() which is overridden for each derived DI container specific Application class it essentially does what you are doing, but is already within the framework.

PR #1608, opened 2017-12-21
Remove prism from master

sibille:


ralarcon:
Both pending builds are executed and passed tests.

Issue #1629, opened 2017-12-31
Master Detail View has bug while setting background as transparent in narrow window mode


hupo376787:
## For new Features:

* **Overview:** What you'd think should be added with scenarios
* **Links:**

## For Bugs:
Master Detail View has bug while setting background as transparent in narrow window mode

### Repro steps

I created a master detail view using latest template, when I resize window to narrow mode(0 - 640), and set controls:MasterDetailsView Background="Transparent", the detail info overlay on the list. See the gif

```
<controls:MasterDetailsView
            Grid.Row="1"
            x:Name="MasterDetailsViewControl"
            Background="Transparent"
            ItemsSource="{x:Bind SampleItems}"
            SelectedItem="{x:Bind Selected, Mode=OneWay}"
            ItemTemplate="{StaticResource ItemTemplate}"
            DetailsTemplate="{StaticResource DetailsTemplate}"
            NoSelectionContentTemplate="{StaticResource NoSelectionContentTemplate}"
            BorderBrush="Transparent" />
```
![1](https://user-images.githubusercontent.com/14960168/34463315-ec3251a2-ee92-11e7-864d-c2c1ae8705ff.gif)

In earlier versions, when there exist "MasterDetailDetailPage", there is no such bug.

### Expected Behavior

When setting controls:MasterDetailsView Background="Transparent", and Click an item, use a page instead, so that the detail info do not overlay on the list.

### Actual Behavior

![1](https://user-images.githubusercontent.com/14960168/34463315-ec3251a2-ee92-11e7-864d-c2c1ae8705ff.gif)

## System

* **VS Version:15.5.2**
* **WTS Wizard Version:1.6.17355.2**
* **WTS Template Version:1.6.17355.2**
* **Windows Build:  16299**

crutkas:
What are your min / target for your project?  Can you share a repro?

crutkas:
@nmetulev have you seen a bug like this in the Toolkit?

hupo376787:
Normally, I set target is 16299, min is 14393. But when I set min is 16299, the bug still exists.
The uwp toolkit is using the same control, so I think it has the same bug when setting background as transparent.
Here is the repo https://github.com/hupo376787/WindowsTemplateStudio---Master-Detail-View-bug.git

mvegaca:
[...]

PR #1630, opened 2018-01-02
Master-Detail control background

mvegaca:
Set up MasterDetail control background as a workaround of Issue #1629


crutkas:
1. update known issues and say we have a workaround.
2. Update bug with workaround
3. Either Toolkit's team or WTS fixes the underling issue.

We shouldn't update our codebase for a bug in their control as anyone that has the page is now affected

Issue #1338, opened 2017-10-23
Improve generated VB code consistency with additional analyzers


mrlacey:
Follow on from #863 

There are a lot fewer code quality analyzers for VB code available than for C#. This means that it is harder to enforce a high level of code standards and consistency.
The SonarLint analysis that we currently do is a good start but we could do more/better.

It would be good to port some of the StyleCop analyzers to support VB projects as a way to standardize more of the generated code.

It would also be good to create some additional analyzers for points highlighted at 
 https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/coding-conventions

Specifics of this issue are still to be defined and it may be something that is done iteratively.

mrlacey:
Moving to current milestone as want to get this done sooner rather than later as any issues may cause problems with future merge support

mrlacey:
I'm implementing this with a combination of additional tests within `Templates.Test` and also a new Roslyn analyzer project hosted separately and distributed via NuGet.

Hosting the other project separately and not making it a part of WTS because:
- Not core to WTS and not something we'd want to commit to supporting in the same way.
- Having it as a separate NuGet package makes it easy to use and also means other can easily use it if they so desire.
- No way for me to publish a new NuGet package from within WTS.

Speak up quickly if there are objections to this.

mrlacey:
Closing as new tests merged to default branch

PR #1657, opened 2018-01-05
Standardize formatting of VB generated code

mrlacey:
For #1338 

Adds some new tests to verify standardizing of VB generated code.  
Some are tests that analyze the template files directly.  
Others are tests that generate the projects and verify with a Roslyn Analyzer. (Like a smaller/simplified version of the C# StyleCop tests.)


mrlacey:
Test that failed is working on my machine. Investigating

mrlacey:
I think the cause of the failing tests (Some vb files not reporting correct line endings on the last line of the file) was due to previous updates (that only added a final CrLf to the file) being ignored.
Not a big issue, so I've disabled the offending test.

PR #1664, opened 2018-01-10
Rename Cross platform poc to WtsXPlat

dgomezc:


crutkas:
why are we having a namespace that is so closely tied to an outcome?  Why isn't it just WTS or WTS.Core?

ralarcon:
Well, that was just a PoC. The intention was to mark in the PoC what we were doing (think that we were playing around with multiple combinations). Now that the model is more or less clear we can use the name we prefer. Not a big deal.

ralarcon:
@dgomezc, when possible, change the Namespace for the PoC to Wts (Wts.Core, Wts.Mobile, Wts.iOS, Wts.Android, Wts.Uwp)

dgomezc:
Rename WtsXPlat to Wts and move poc to pocs branch (#1675).

PR #1683, opened 2018-01-12
Fix emulator when not content available

dgomezc:


crutkas:
where is the associated bug here?  What caused this?

ralarcon:
Discovered as part of the optimizations just after merging to dev. That's me asking for a quick fix...

Issue #1504, opened 2017-11-29
MediaPlayer Page don't Pause the video on Tabbed project type


mvegaca:
When you navigate to a MediaPlayer page on Tabbet project the video starts because it's configured as auto-start, but when you navigate to another page the video does not be paused because Pause action is set on OnNavigatedFrom event handler and this event is not fired on tabbed project

dgomezc:
After reviewing this issue, we think we will need some infrastructure for Pivot project types to handle when the active tab changes and properly execute actions (in this case play / pause the video when required). 

Then, the action or preference itself would depend on the developer needs, that is some apps would require the video to be paused and other would prefer the background playing... 

I've opened a new issue to discuss the convenience of having some infrastructure to detect when the active tab have changed: #1584

Considering all of this, I would suggest to hold on this until decide if we will have the required infrastructure.

ralarcon:
Moving to next release as we are not going to face this for 1.6

crutkas:
@dgomezc  Just verifying that we have tested this on navview along with tabbed?  code has been removed from the main page so just want to be sure the page still works correctly in the majority use case.

sibille:
Verified in dev-nightly:
Wizard: 0.16.18018.65001
Templates: 0.16.18018.1

PR #1696, opened 2018-01-15
Fix Media Player page in Caliburn Micro Framework and Pivot project.

dgomezc:
## PR checklist

Quick summary of changes

#1504

### DIFF
--- a/templates/_composition/CaliburnMicro/Page.MediaPlayer.AddNavigation_TabbedPivot/.template.config/template.json
+++ b/templates/_composition/CaliburnMicro/Page.MediaPlayer.AddNavigation_TabbedPivot/.template.config/template.json
@@ -3,13 +3,13 @@
   "classifications": [
     "Universal"
   ],
-  "name": "Page.MediaPlayer.AddNavigation_TabbedPivot",
+  "name": "CaliburnMicro.Page.MediaPlayer.AddNavigation_TabbedPivot",
   "tags": {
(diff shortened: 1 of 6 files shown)


ralarcon:
@mrlacey we have changed the VB implementation too. The change was simple: to use a different method from the toolkit (`FindAscendant`) to ensure we find the Pivot page where the media player page is. We requested your review just as a double check.

ralarcon:
@mrlacey Just double checked (on both @dgomezc and me laptops) and it worked fine. 

Can you double check that the latest templates were correctly loaded to the emulator? The generated project should have the "FindAscendant / FindDescendant" method calls.

Now, to the templates being refreshed on VSEmulator or Installer (vsix running locally on the experimental instance) you would need to refresh the template engine chache (not done by default to speed up the wizard loading). You can do it by clicking "Refresh" button in the VSEmulator or, from the wizard, using a new button available in debug complilation mode which appears in the same place than the wizard / templates version numbers (clicking in the upper-left corner button).

mrlacey:
My mistake. I was looking at the [wrong branch](https://github.com/dgomezc/WindowsTemplateStudio/tree/Issue1504-PivotAndMediaPlayerPage). 

Issue1504-PivotAndMediaPlayerPage and Issue1504-MediaPlayerPivotFix :/

ralarcon:
No problem at all! Double checked 👍

Issue #1593, opened 2017-12-19
update roadmap


crutkas:
Roadmap shows 1.3 plans when we are looking at 1.7 and 2.0

mrlacey:
I think it will be beneficial to look at the whole roadmap and milestone/release process as part of updating the roadmap document.

At the moment, the process seems very disorganized and unstructured.

Here's how it seems the release occurs (based on the last few milestone releases)
- A milestone/release number if defined with no deadline/due date.
- Issues are assigned to the milestone on a seemingly random basis.
- There is a rush of work on some items assigned to the milestone early in its existence.
- Work on items assigned to the milestone slows after a couple of weeks.
- Entirely new issues are created and assigned to the milestone and are worked on straight away.
- Some items are moved to a future milestone without warning or explanation.
- There's a rush of work on the remaining issues and then a release is made with very little time for testing.
- The next [minor] release has new issues assigned to it without any indication of why and so that release becomes a mix of assorted new issues and issues carried over from the last release. (Some issues end up getting carried over several milestones.)


There are a number of issues here that I believe could and should be improved by following these action points

- Set a deadline for each milestone/release. It can, of course, be adjusted if needed but without one there's less incentive to get work that is part of a release done. It also helps with wider planning goals and external visibility and expectations. Particularly when someone is waiting for a particular feature of bug fix.
Deadlines for further out releases can be less specific and are more likely to change but upcoming minor releases should be, as far as possible, fixed and accurate.
At the moment there is a planned major update (2.0) but there is no way to know when this might be coming, what it might contain, or how many minor updates may happen before it comes.
A deadline also makes it possible to assess how realistic it will be to complete the amount of work that can be performed as part of a release. Without a deadline, there's always room to add more issues. With a deadline, it makes it possible to
The alternative to not having a deadline is to strictly define a set of issues to be completed and then release when they are done and tested. This way of working means that new items can't be added to a release as doing so extends the deadline.
Yes, there's a balance to be found but making it up as you go along isn't the answer. Ideally, the goal should be to have a set of features/issues that can be confidently completed in the amount of time defined. Doing this requires a clear definition of what is required for each feature before it is assigned to a milestone--which is a separate conversation that needs to be had.

[...]

PR #1721, opened 2018-01-19
Updated roadmap (& standardized formatting)

mrlacey:
For #1593
Updated with discussed areas for future development.

@crutkas feel free to modify as you feel necessary. Just raised this to make sure it doesn't get forgotten or overlooked.


crutkas:
Good with it. Looking at source. The dot in VB.Net has an encode VS a period.

crutkas:
Also thanks for updating :)

mrlacey:
> The dot in VB.Net has an encode VS a period.

That's to stop GitHub/Markdown interpreting it as a URI.

PR #1725, opened 2018-01-23
Add prism and vb back into master

sibille:
## PR checklist

Quick summary of changes
Added prism and vb back into master


sibille:
Both required builds ran correctly

Issue #1729, opened 2018-01-24
Get templates from package delivered with the vsix on first use of wizard


sibille:
This is done only when there is no content available for the wizard. 
In case there was a previous version of the wizard compatible with the current wizard (f. ex version 0.11.0.1) those templates are used and 0.11.0.2 templates are downloaded in background.

0.11.0.2 templates should be extracted from the package delivered with the vsix.

sibille:
Verified in dev-nightly:
Wizard: 0.16.18025.3
Templates: 0.16.18025.3

PR #1736, opened 2018-01-24
Remove 65000 from dev wizard version

sibille:
## PR checklist

Quick summary of changes
In dev wizard version is 65000 + revision number, but templates just have revision number.
Removing this to align with other environments. 
#1729

### DIFF
--- a/_build/Extension-SetIdentityAndVersion.ps1
+++ b/_build/Extension-SetIdentityAndVersion.ps1
@@ -43,9 +43,9 @@ if($buildNumber -match $VersionRegEx){
 
   $revision =  [int]::Parse($matches[4]).ToString()
 
-  if($buildNumber.ToLower().StartsWith("dev")){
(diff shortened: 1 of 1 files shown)


_build/Extension-SetIdentityAndVersion.ps1
```diff
@@ -43,9 +43,9 @@ if($buildNumber -match $VersionRegEx){
 
   $revision =  [int]::Parse($matches[4]).ToString()
 
-  if($buildNumber.ToLower().StartsWith("dev")){
-    $revision = (65000 + [int]::Parse($matches[4])).ToString();
-  }
+  #if($buildNumber.ToLower().StartsWith("dev")){
```
mrlacey: Why has this been kept but commented out? Why not just delete it?

sibille: Your right, going to delete

PR #1967, opened 2018-02-22
Update TextBlock Styles in project templates

mvegaca:
Update TextBlock Styles in project templates


mrlacey:
Be careful with the naming of commits. This only changes XAML whitespace. It's not changing (updating) the styles at all.

Why is this being done without an issue raised first?
Is consistency of XAML formatting through all template files being considered? If not, why just do this? If it is then that's a potentially big change that may need lots of consideration and then new tests and processes to ensure the formatting style is maintained.

mvegaca:
Sorry Matt, I forgot to create a new Issue!

mvegaca:
Here is the Issue

Issue #1555, opened 2017-12-15
Prism: Reduce use of composition templates where not necessary


sibille:
The following prism feature/pages have composition templates for prism that are not necessary as the feature/page has a separate template for Prism:

- ShareTarget, 
- Chart, 
- Grid
- MasterDetail, 
- ImageGallery 

Intention behind composition template was to augment reusibility of code. If neither the page/feature template nor the composition template are shared their content should be merged in the page/feature template to reduce complexity.

crutkas:
is it possible to validate stability and get this in to the 1.6 release?  should this be a 1.7?

ralarcon:
Not to much risk as I understand but there is no hurry. This falls in the side of template optimization, would say, we can leave it to 1.7 if needed.

sibille:
This is not much risk, @ralarcon, @crutkas should we do this for 1.7?

sibille:
ShareTarget Feature is missing extension in appxmanifest and resources are not replaced correctly.

sibille:
Verified in dev-nightly:
Templates version: 0.18.18114.1
Wizard version: 0.18.18114.1

PR #2000, opened 2018-02-27
Issue1555 reduce prism compostion templates

milazzom:
This PR addresses issue #1555, reducing composition templates for the Prism framework where applicable.  No new features were added, therefore no new automated tests are included in this PR.


templates/Features/SampleData.Prism/Services/ISampleDataService.cs
```diff
@@ -0,0 +1,14 @@
+using Param_ItemNamespace.Models;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Threading.Tasks;
+
+namespace Param_ItemNamespace.Services
+{
+    // This class holds sample data used by some generated pages to show how they can be used.
```
mrlacey: This comment doesn't apply to the interface as it does to the concrete clss

templates/Features/SampleData.Prism/Services/ISampleDataService.cs
```diff
@@ -0,0 +1,14 @@
...
+    // TODO WTS: Delete this file once your app is using real data.
```
mrlacey: Shouldn't this TODO comment be specific to this interface

templates/Features/ShareTarget.Prism/ViewModels/ShareTargetFeatureViewModel.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Param_ItemNamespace.ViewModels
 {
-    public class wts.ItemNameViewModel : BindableBase
+    public class ShareTargetFeatureViewModel : BindableBase
```
mrlacey: This seems like a bad name. Why switch from the one based on the template item name?

milazzom: The sourceName in the template file is ShareTargetFeature, so that would be replaced at generation time instead of wts.ItemName.

mrlacey:
PR also includes a failing unit test relating to file encoding.

milazzom:
@mrlacey I left some of the items in the composition templates to prevent dead code from being introduced into the sample data service feature, or augment files that are generated or referenced from other templates.  For example, if I add a chart page, I don't want to include the models/methods that are only used by the image gallery in the sample data service.  Additionally, the Resources file is generated by other templates.  If it exists, I want it updated by the composition templates based on what the user added.  Is it possible to have post-actions placed directly in the templates themselves as opposed to storing them in a composition template?  I think if the answer is yes, then I can move a lot of the postactions into the templates themselves.  Otherwise, I went with convention seen in other templates.

sibille:
@milazzom Yes, you can have _postaction files in any template. Postactions and composition templates are independent pieces, often used together but not necessarily. A composition template is used to avoid duplication of code in templates. Lets say Template A and Template B need to do the same thing (file creation or file edition via postaction), you'd put this code in a composition template C and have it triggered by A and B.

milazzom:
@sibille Thank you for the clarification.  
@mrlacey I'll resubmit after implementing your feedback.  Thanks!

milazzom:
@mrlacey I think this PR is ready for re-review.  I ran the PS script to fix template encoding, and eliminated a good number of the composition templates based on our previous conversation.  When running the unit tests, I get a lot of OutOfMemoryExceptions for the template build tests, but I did verify that the file encoding tests failed, then passed after running the script.  Thanks!

sibille:
@milazzom, where do you get OutOfMemory exceptions executing tests? 
Running the Execute-LocalTestsInParallel.ps1 script? We had this issue on build machines due to parallelizing too many traits, and solved it splitting the run in two blocks, I'll apply the same change to the script. I rerun your PR on appveyor, and there are still some tests failing with param errors, I'll have a look at it to give you more details

milazzom:
Hi @sibille, I was seeing OutOfMemoryException occur when running the tests in GenProjectTests.cs.  I'll checkout the AppVeyor output and adjust.  Thanks!

sibille:
@mrlacey, I think all issues are fixed here, could you doublecheck and adapt your review to be able to merge this?

mrlacey:
@sibille I trust your review ;) -- i.e. I've haven't rechecked

Issue #2042, opened 2018-03-09
Uri Scheme doesn't include de Microsoft.Toolkit.Uwp License


mvegaca:
UriScheme is including the toolkit in this [postaction](https://github.com/Microsoft/WindowsTemplateStudio/blob/dev/templates/Features/UriScheme/_postaction.csproj) but is not including that as a template license.

![image](https://user-images.githubusercontent.com/5556784/37195614-e8c4c408-2373-11e8-8fde-cbb95e608d5e.png)

mvegaca:
The same problems happens in MediaPage when it is added on a Pivot project Type and this License is not added in media player page or pivot project type.

crutkas:
Wait, what?  why does it include the toolkit?  It feels unnecessary for what the feature does.

crutkas:
Further more.  Where is the developer todo to rename the protocol?

mvegaca:
I've checked that Toolkit is not used on Uri Scheme, we will remove the post action that adds the nuget reference.

The rename TODO comment for update the UriScheme protocol is in SchemeActivationHandler file line 10. No changes here.

mvegaca:
Add Microsoft.Toolkit.Uwp license on TabbedPivot-MediaPlayerPage composition templates #2043

mvegaca:
Remove Microsoft.Toolkit.Uwp.Notifications from UriScheme Templates #2043.

sibille:
Verified in dev-nightly: 
Templates version: 0.17.18071.1
Wizard version: 0.17.18068.1

PR #2043, opened 2018-03-09
Add Microsoft.Toolkit.Uwp license on TabbedPivot-MediaPlayerPage comp…

mvegaca:
Uri Scheme doesn't include de Microsoft.Toolkit.Uwp License #2042


mrlacey:
The title here says to add a license, but the commit comment is to remove the reference. Seems unclear what's happening here.

Issue #962, opened 2017-08-16
Add documentation of SettingsStorageExtensions


mrlacey:
_Based on [this](https://github.com/Microsoft/WindowsTemplateStudio/issues/918#issuecomment-322631160)_

Add documentation to explain the use of and reasons for the Settings extensions and functionality.
Include:
- Why it exists
- Why async (& how to use without async)
- Use with different types
- How and why may not want to use it

PR #2048, opened 2018-03-12
Add c# and VB setting storage documentation.

dgomezc:
## PR checklist

Quick summary of changes

#962 - Add documentation of SettingsStorageExtensions

### DIFF
--- a/docs/features/setting-storage.md
+++ b/docs/features/setting-storage.md
@@ -1,8 +1,8 @@
-#Setting Storage Extensions
+# Setting Storage Extensions
 
 SettingsStorageExtensions is a helper class that groups extension methods to facilitate the use of serialization, handling and storing of any content in a UWP application. This extension wraps the handling and storing of content (for example: custom content, key-values, files, etc.) in the same way as simple settings for a consistent interface.
 
-##Advantages:
+## Advantages:
  - All methods of handling and storing content can be found in the same class and using a similar nomenclature. 
  - Facilitate the serialization and deserialization of data content and make it transparent for the developers.
  - Using methods that avoid problems in the serialization process (for example: big files that can take a long time to process).
@@ -16,7 +16,7 @@ This helper contains extension methods for handling and storing content in:
 More info: [https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data](https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data).
 
 
-##Examples of use:
+## Examples of use:
 
 To use these extension methods you have to add a reference to:
 ```csharp
@@ -31,7 +31,7 @@ public class MyData
     public string Name { get; set; }
 }
 ```
-###Application Data
+### Application Data
 
 - Save to ApplicationData: saves MyData object as a key-value pair where the value is serialized in json.
 
@@ -56,7 +56,7 @@ private async Task<MyData> ReadMyDataAsync()
     return await ApplicationData.Current.LocalSettings.ReadAsync<MyData>(key);
 }
 ```
-###Storage folder
+### Storage folder
 - Save content to StorageFolder: saves the serialized MyData object in json to a file in the StorageFolder.
 ```csharp
 private async Task SaveMyDataAsync()
--- a/docs/features/setting-storage.md
+++ b/docs/features/setting-storage.md
@@ -1,0 +1,113 @@
+#Setting Storage Extensions
+
+SettingsStorageExtensions is a helper class that groups extension methods to facilitate the use of serialization, handling and storing of any content in a UWP application. This extension wraps the handling and storing of content (for example: custom content, key-values, files, etc.) in the same way as simple settings for a consistent interface.
+
+##Advantages:
+ - All methods of handling and storing content can be found in the same class and using a similar nomenclature. 
+ - Facilitate the serialization and deserialization of data content and make it transparent for the developers.
+ - Using methods that avoid problems in the serialization process (for example: big files that can take a long time to process).
+ - All methods automatically give default values back if the key you're trying to read doesn't exist.
+ - It makes working with byte arrays as easy as working with non-binary data (which can sometimes be complicated.)
+
+This helper contains extension methods for handling and storing content in:
+ - StorageFolder
+ - ApplicationData
+
+More info: [https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data](https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data).
+
+
+##Examples of use:
+
+To use these extension methods you have to add a reference to:
+```csharp
+using MyApp.Helpers;
+```
+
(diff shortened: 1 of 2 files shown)


docs/features/setting-storage.md
```diff
@@ -0,0 +1,116 @@
+# Setting Storage Extensions
+
+:heavy_exclamation_mark: There is also a version of [this document with code samples in VB.Net](./setting-storage.vb.md) :heavy_exclamation_mark: |
+----------------------------------------------------------------------------------------------------------------------------------------------- |
+
+SettingsStorageExtensions is a helper class that groups extension methods to facilitate the use of serialization, handling and storing of any content in a UWP application. This extension wraps the handling and storing of content (for example: custom content, key-values, files, etc.) in the same way as simple settings for a consistent interface.
```
mrlacey: long, wordy, and unclear

mrlacey: Multiple extensions, not a single extension.

docs/features/setting-storage.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Advantages:
+ - All methods of handling and storing content can be found in the same class and using a similar nomenclature. 
+ - Facilitate the serialization and deserialization of data content and make it transparent for the developers.
+ - Using methods that avoid problems in the serialization process (for example: big files that can take a long time to process).
+ - All methods automatically give default values back if the key you're trying to read doesn't exist.
+ - It makes working with byte arrays as easy as working with non-binary data (which can sometimes be complicated.)
+
+This helper contains extension methods for handling and storing content in:
```
mrlacey: Need to explain the difference between these. The code examples below show no difference between the two.

docs/features/setting-storage.md
```diff
@@ -0,0 +1,116 @@
...
+ - StorageFolder
+ - ApplicationData
+
+More info: [https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data](https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data).
```
mrlacey: For additional reference, it would be good to mention the `StorageFileHelper` in the toolkit and highlight how these are different and complimentary.

docs/features/setting-storage.md
```diff
@@ -0,0 +1,116 @@
...
+- Get content from StorageFolder:
+```csharp
+private async Task<MyData> ReadMyDataAsync()
+{
+    var key = "MyDataKey";
+    return await ApplicationData.Current.LocalFolder.ReadAsync<MyData>(key);
+}
+```
+
+- Save file to StorageFolder: Save file in the StorageFolder.
```
mrlacey: duplication?

docs/features/setting-storage.md
dgomezc: New proposal:

 SettingsStorageExtensions is a helper class that groups extension methods to facilitate the use of serialization, handling and storing of any content in the same way as simple settings for a consistent interface (for example: custom content, key-values, files, etc.).

docs/features/setting-storage.md
dgomezc: Between StorageFolder and ApplicationData?

docs/features/setting-storage.md
dgomezc: Do you refer to duplicate method names?

docs/features/setting-storage.md
dgomezc: I´m not sure if this adds value to our documentation. It could cause confusion, as this helper is not a part of our generated code.

docs/features/setting-storage.md
dgomezc: See new proposal.

docs/features/setting-storage.md
```diff
@@ -0,0 +1,116 @@
...
+```csharp
+private async Task SaveMyFileAsync()
```
mrlacey: I think this example could use some more comments to make it clearer what's being shown and what's specific to the helper.

docs/features/setting-storage.md
```diff
@@ -0,0 +1,116 @@
...
+ - ApplicationData
+
+More info: [https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data](https://docs.microsoft.com/es-es/windows/uwp/design/app-settings/store-and-retrieve-app-data).
+
+
+## Examples of use:
+
+To use these extension methods you have to add a reference to:
+```csharp
+using MyApp.Helpers;
```
mrlacey: Format the code like below to be consistent with other documentation
`using {YourAppName}.Helpers;`

docs/features/setting-storage.md
```diff
@@ -0,0 +1,116 @@
...
+### Application Data
+
+- Save to ApplicationData: saves MyData object as a key-value pair where the value is serialized in json.
+
+```csharp
+private async Task SaveMyDataAsync()
+{
+    if (ApplicationData.Current.IsRoamingStorageAvailable())
+    {
+        var key = "MyDataKey";
```
mrlacey: Is it really necessary to define this 4 times? 
I think it shows a bad practice.

docs/features/setting-storage.md
sibille: ok, good point!

docs/features/setting-storage.md
sibille: ok

sibille:
@mrlacey, we were not sure if we understood correctly some of your comments before reworking the text, thats why we put comments on you comments, but this is getting a bit confusing, so let me try to resume: 
- When you say we need to explain the difference between the two are you refering to StorageFolder and ApplicationData? I think that makes sense, but the comment is pointing to another line.
- When you say duplication do you refer to the method names or to the description?
- MyDataKey is defined 4 times to make the examples independent of each other.
- What other points are you refering to that are not covered by this document?

docs/features/setting-storage.md
mrlacey: The text in line 86 seems to say the same things twice.
 "Save file to StorageFolder: Save file in the StorageFolder."

docs/features/setting-storage.md
mrlacey: Yes, between StorageFolder and ApplicationData?

mrlacey:
> MyDataKey is defined 4 times to make the examples independent of each other.

They don't need to be independent. Showing a const being defined inline of the method where it's used is a bad practice and shouldn't be encouraged. Working on the maxim that "All code on the internet will be cut and pasted into someone's production code." we shouldn't show something that we wouldn't want to see in anyone's production code.


> What other points are you refering to that are not covered by this document?

The original issue stated that the documentation should include:

- Why it exists
- Why async (& how to use without async)
- Use with different types
- How and why may not want to use it

The original purpose of adding this documentation was designed to answer questions like those asked [here](https://github.com/Microsoft/WindowsTemplateStudio/issues/918#issuecomment-322631160).

sibille:
@mrlacey, we reviewed the doc and incorporated your suggestions. If you want to add sth in, emphasis or change anything feel free to edit.

PR #2066, opened 2018-03-14
Remove unnecessary `Exit Select` calls

mrlacey:
Minor tweak to VB templates to remove unnecessary code.


sibille:
@mrlacey, this has conflicts due to Navview updates, could you have a look?

Issue #2093, opened 2018-03-26
Make sample data platform agnostic


mrlacey:
Remove Windows specific references from `SampleOrder` and `SampleDataService`.

This is to enable the use of sample data in multiple projects.
X-Ref #2087

sibille:
There is a wrong char in Prism.SampleDataService.

mrlacey:
Closing as verified in 
   wizardVersion="v0.18.18112.1" 
    templatesVersion="v0.18.18112.1"

Prism issue appears to be due to https://github.com/Microsoft/WindowsTemplateStudio/commit/1758a330d9a57e5a2aa5cce068c5a38a9313fd0c#diff-fc87c59e6a12d914f43fe8c63973a9c6

PR #2096, opened 2018-03-27
Make sample data X-Plat

mrlacey:
For #2093 
Removing Windows-specific types from classes in advance of potentially moving them to a separate project.


mrlacey:
Failure is due to being based on dev but needing to merge into v2.1.
I will update and retarget to remove conflicts

Issue #916, opened 2017-08-08
Need the ability to add an extra project to the solution as part of features


mrlacey:
Need the ability to add an additional project to the generated solution.
As part of a feature, the template would need the ability to be able to add to and modify (postaction-merge) both the original project and the new one.

Implications for Right-Click>Add need to be explored and understood for this before implementing. Features that want to add extra projects shouldn't be prevented from being included as right click options.

Working on the assumption that the added project would be a library or set of tests and so couldn't later be extended by having extra things added to them through rightclick-add.
Any features that add another project to the solution would exist in isolation and so we wouldn't support features adding things to secondary projects. Unless there's a good reason why this shouldn't (at least initially) be the case.

This is needed by features which will add testing capabilities to the generated code (X-Ref #299 & #300) and possibly more.

Xamarin PoC has the ability to produce a solution with multiple projects in the generated solution but not explored if this can be done optionally. I presume this would need

mrlacey:
For reference.
There is a way to do this as a postaction in the config file but this is for the generation of new projects, not the addition of one based on a defined template.

Docs: https://github.com/dotnet/templating/wiki/Post-Action-Registry#add-a-reference-to-a-project-file
Example: https://github.com/dotnet/templating/issues/601

I have looked at the Xamarin PoC branch but that just shows generating a multi-project solution. It doesn't allow adding a project to a solution through composition.

mrlacey:
This needs to: 
- add the new project to the solution
- Have the new project add a reference to the existing project (and so will need the name of the existing project)
- Update code files in the new project based on the name of the existing project.

All this at least in the initial version.

In future, we'll also need to be able to:
- Have existing project reference the newly added one
- Make changes to the existing project based on the name of the new one (and other classes, etc. in it)
- Allow for specifying the name of the added project (possibly)

Shouldn't allow for circular referencing of added project. Reference should be one way or the other.

mrlacey:
Removing myself as assignee as don't want to block others from moving this forward.

mrlacey:
Moving to backlog as this is only needed to unblock #299 & #300 and both of them are low priority (for now)

crutkas:
This should be there with xamarin as well

mrlacey:
[...]

PR #2108, opened 2018-03-30
Ability to have features that add a project to the solution

mrlacey:
For #916 
Extends v2.1 branch to better support multi-project solutions.
Adds `TemplateType.ProjectFeature` so a feature can add a new project to the solution. (The UI treats the template as a feature but the engine treats it as a project. - It's not possible to have a template that does both but the same result can be achieved with a feature template that does one part, such ass add the new project, and a composite template that depends on it and modifies the main app to reference the new project.)
This is necessary for features relating to testing, data access, and X-Plat support.

This is just the capability to have features that add projects to the solution, for examples of having such features see https://github.com/mrlacey/WindowsTemplateStudio/tree/Issue300-UiTestProject and https://github.com/mrlacey/WindowsTemplateStudio/tree/MiscProjectFeatureTests


mrlacey:
@sibille I was using wts.type as we have logic that uses that for selection of actions to execute and what goes where in the UI. A change here could be big and far reaching. I don’t want to make a potentially big change if it doesn’t give us some benefit. Have you looked into what else we could get form he template engine here?

sibille:
@mrlacey, I was referring to use both, wts.type for UI things and type for postactions. The change should not be too big, but I didn't try to do it. I'll set up a poc tomorrow.

sibille:
@mrlacey, sorry for the delay, I just set up the poc on https://github.com/sibille/WindowsTemplateStudio/tree/poc_projectfeature. It's just a quick test, we'll have to find good names for the two types in templates (one is coming from template engine, the other is ours)

Issue #2140, opened 2018-04-05
Dev-nightly: Exception when navigating to detail of ImageGallery in NavView


sibille:
1. Create new SplitView project
2. Add Image Gallery
3. Build and start app
4. Go to imageGallery
5. Try to navigate to detail

-> Exception: 
{System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at App16.Views.ShellPage.Frame_Navigated(Object sender, NavigationEventArgs e)
   at Windows.UI.Xaml.Navigation.NavigatedEventHandler.Invoke(Object sender, NavigationEventArgs e)
   at App16.Services.NavigationService.Frame_Navigated(Object sender, NavigationEventArgs e)}

mvegaca:
Fixed on #2142

sibille:
Verified on dev-nightly: 
Templates version: 0.18.18114.1
Wizard version: 0.18.18114.1

PR #2142, opened 2018-04-05
Fix Issue on NavigationView Frame Navigated

mvegaca:
Fix an issue when ImageGallery navigates on a nav pane project


sibille:
fixes #2140

Issue #2145, opened 2018-04-06
Wrong image restore on go back navigation


mvegaca:
Repro steps:
 - Create an app with nav view with pages: Blank, Image Gallery and Settings.
 - Launch the app.
 - Navigate to the image gallery.
 - Navigate to the first image.
 - Move the selected image to the next image in the pivot.
 - Navigate to the settings page.
 - Click on go back.

Expected behavior:
Go back to the image detail page on the second image.

Actual behavior:
Go back to the image detail page on the first image.

crutkas:
This feels like a must fix for launch

mrlacey:
Note. this is an existing issue in the current code. Not a regression after changing to the NavigationView as I initially assumed when seeing this raised.

sibille:
I think we can finish it today

sibille:
Verified on dev-nightly:
Templates version: 0.18.18114.1
Wizard version: 0.18.18114.1

PR #2148, opened 2018-04-09
Issue2145 imagegallery

mvegaca:
Fix Wrong image restore on go back navigation #2145


mvegaca:
Could you please @mrlacey review the VisualBasic updates?

Issue #2262, opened 2018-05-18
Build dev.templates.tests.full_20180518.2 failed


wasteam:
## Build dev.templates.tests.full_20180518.2
- **Build result:** `failed`
- **Build queued:** 5/18/2018 6:21:32 AM
- **Build duration:** 23.58 minutes
### Details
Build [dev.templates.tests.full_20180518.2](https://winappstudio.visualstudio.com/web/build.aspx?pcguid=a4ef43be-68ce-4195-a619-079b4d9834c2&builduri=vstfs%3a%2f%2f%2fBuild%2fBuild%2f25675) failed

+ xunit.console.exe :     GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "SplitView", framework: 
"CodeBehind", platform: "Uwp") [FAIL]
At pbatch:27 char:27
+ 
    + CategoryInfo          : NotSpecified: (    GenerateAll...: "Uwp") [FAIL]:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
    + PSComputerName        : [localhost]
 
    GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "SplitView", framework: "MVVMLight", platform: 
"Uwp") [FAIL]
    GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "SplitView", framework: "MVVMBasic", platform: 
"Uwp") [FAIL]
    GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "Blank", framework: "CodeBehind", platform: 
"Uwp") [FAIL]
    GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "Blank", framework: "MVVMLight", platform: 
"Uwp") [FAIL]
    GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "Blank", framework: "MVVMBasic", platform: 
"Uwp") [FAIL]
    GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "TabbedPivot", framework: "CodeBehind", 
platform: "Uwp") [FAIL]
    GenerateAllPagesAndFeaturesAndCheckWithSonarLintAsync(projectType: "TabbedPivot", framework: "MVVMLight", 
platform: "Uwp") [FAIL]
[...]

PR #2263, opened 2018-05-18
Fix tests

mvegaca:


mrlacey:
For a PR like this, it would be helpful to have a more descriptive title and some more information in the description. What tests are failing? Is this related to something new?--if so, why a new PR and not just an addition to an existing PR?

This also raises the question of why code that causes tests to fail got merged into the dev branch. Is this something we could/should have caught earlier?

sibille:
It's related to #2262, VB templates were failing due on VB Analysis tests. This cannot be detected in Appveyor, as VB projects cannot be executed here. We use to execute VB tests on local machine to avoid this until we find a better solution.

PR #2267, opened 2018-05-18
Update xamarin poc

dgomezc:


mrlacey:
Is there anything here other than a namespace change?

dgomezc:
No. The poc was updated with a new name that was more representative, but it was only done in the xamarin-infra branch. Now we have removed the xamarin pocs from xamarin-infra and dev branches, and updated the name in the pocs branch.

The objective is to have all pocs only in the pocs branch, in their latest version.

PR #2313, opened 2018-05-25
Fix 589534

mvegaca:
Accessibility Issue 589534 Change delete button color in high contrast


mrlacey:
What is "589534" if a reference to an ID in another system, can they be linked or at least duplicated so that it's possible to asses the changes made against what was originally raised as the issue?

mvegaca:
Is an issue reported by accessibility tool and this issues are not tracked on GitHub but I need to add the ID in the commit, it's not necessary to create a new Issue on GitHub.

sibille:
@mrlacey, The issue is about the delete button having the wrong foreground color in high contrast mode, should be black, not white

image

mrlacey:
It's not just this issue I was referring to, I've seen other PRs created like this. 
Having the complete history of a change in GitHub, where it's public, helps with transparency and tracking all work in progress in a single place.

PR #2369, opened 2018-06-19
Corrects error message.  I found this in our STS fork.

milazzom:
This PR corrects the error message when an invalid wts.group value is encountered.  We added a new Telemetry group in our STS fork, and were seeing this error incorrectly reported when running the unit tests.  

This PR does not change or add any functionality.


sibille:
Thanks @milazzom

Issue #2387, opened 2018-06-22
WindowManagerService does not apply current Theme to new Windows


mvegaca:
1. Create a new project with a blank page, a settingsPage and MultiView feature 
2. Add code to open a new window following this [doc](https://github.com/Microsoft/WindowsTemplateStudio/blob/master/docs/features/multiple-views.md).
3. Start the app 
4. Go to settings and change theme to Dark
5. Open a new window using the MultiView feature

Expected behaviour: 
The new window should show in dark theme.

Actual behaviour: 
The new Window should show in light theme. 

We need to pass theme as Parameter and set in the new window frame.

```
public async Task<ViewLifetimeControl> TryShowAsStandaloneAsync(string windowTitle, Type pageType)
{
    var theme = (Window.Current.Content as FrameworkElement).RequestedTheme;
    ViewLifetimeControl viewControl = await CreateViewLifetimeControlAsync(windowTitle, pageType, theme);
    SecondaryViews.Add(viewControl);
    viewControl.StartViewInUse();
    var viewShown = await ApplicationViewSwitcher.TryShowAsStandaloneAsync(viewControl.Id, ViewSizePreference.Default, ApplicationView.GetForCurrentView().Id, ViewSizePreference.Default);
    viewControl.StopViewInUse();
    return viewControl;
}

private async Task<ViewLifetimeControl> CreateViewLifetimeControlAsync(string windowTitle, Type pageType, ElementTheme theme)
{
    ViewLifetimeControl viewControl = null;

    await CoreApplication.CreateNewView().Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () =>
    {
        viewControl = ViewLifetimeControl.CreateForCurrentView();
        viewControl.Title = windowTitle;
        viewControl.StartViewInUse();
        var frame = new Frame() { RequestedTheme = theme };
        frame.Navigate(pageType, viewControl);
        Window.Current.Content = frame;
        Window.Current.Activate();
        ApplicationView.GetForCurrentView().Title = viewControl.Title;
    });

    return viewControl;
}
```

crutkas:
shouldn't this grab the theme by the service, not be passed in?

mvegaca:
I could retrieve it from theme service but if MultiView is added by right click we don't know if ThemeService is added or not.

mrlacey:
I have three concerns about this issue.

**Concern 1 - The repro is inadequate.** 
[...]

PR #2400, opened 2018-06-26
Add RequestedTheme in MultiView new Windows

mvegaca:
WindowManagerService does not apply current Theme to new Windows #2387


sibille:
Depends on doing #2403, will update when this is there

Issue #2360, opened 2018-06-09
Iterating between Maps / Camera


Pinox:
I have noticed that iterating very fast between the maps and camera page causes the app to crash.

VS 15.7.3 Stable
Windows 10 Pro - Version 1803

![ts_version](https://user-images.githubusercontent.com/5297255/41191407-3a63fd42-6bef-11e8-8f1a-a4921b42802d.PNG)

![ts](https://user-images.githubusercontent.com/5297255/41191408-3c767a06-6bef-11e8-820f-0a4029aaad78.PNG)

crutkas:
Can you provide a stack on the crash?

crutkas:
@sibille this feels like the camera hasn't fully shut down and is still in use when a use would bring it back up.  Do we have logic to catch this?

sibille:
We'll have a look at it.

sibille:
@Pinox, which framework did you use to create the application?

Pinox:
@sibille  , I used the latest framework.  Looks like the crash happens when the init of the camera is still busy.    I flagged this as I can see this being an bigger issue on lower specked devices/pc's or slower device drivers where the camera is still busy and trying to initiate.

sibille:
Thanks @Pinox !

dgomezc:
The problem is that the camera resource is initialized and released in the Loaded and Unloaded events of the CameraControl control.

There are cases in which the Loaded event of the control of the page to which we navigate is triggered, before the Unloaded event from the page from which we are navigating, so it tries to initialize the camera resource before releasing it.

To fix this, it is necessary to release the camera resource during navigation in the OnNavigatingFrom event and initialize the camera resource in the OnNavigatedTo event.

In Pivot type projects, these events are not triggered when changing pivot, so we must subscribe to the SelectionChanged event of the pivot, and initialize or release the camera depending on the pages involved in the navigation.

To simplify the code, we have thought of introducing a Helper for this project type, which is initialized when the application is started, receives the pivot, subscribes to the pivot SelectionChanged event, filters the pages involved in the change of the tab and launches an own event indicating those pages. 

A simplified version of the helper would be like this:

```csharp
public class PivotNavigationHelper
{
    private static Pivot _pivot;
    public static event EventHandler<PivotNavigationEventArgs> OnPivotNavigation;

    public static void Initialize(Pivot pivot)
    {
        _pivot = pivot;
[...]

PR #2416, opened 2018-06-29
Issue2360 pivotpages

mvegaca:
Iterating between Maps / Camera #2360


mvegaca:
Hi @mrlacey 
I've updated VisualBasic templates to include the Pivot Behavior.
There is a failing test on VisualBasic named **EnsureVisualBasicCodeDoesNotIncludeCommonPortingIssues** that checks Async on methods with As Task declaration. It's is failing on the IPivotPage Interface but I can't fix it because Interfaces can not declare Async Methods. Could you help me reviewing the test implementation?

sibille:
@LeeParrishMSFT  or @milazzom, could you have a look at the prism changes?

mrlacey:
@mvegaca I added appropriate entries for the exceptional circumstances here and tests now pass

sibille:
thanks @mrlacey

Issue #2501, opened 2018-08-23
Multi view fails with CaliburnMicro


mvegaca:
### Repro steps
Create a C# or VB app with SpliView and CaliburnMicro
Add Main, Blank and Settings pages
Add multiview feature
Add a button in the main page with the WindowManagerService call to open the blank page in a new window.

### Expected Behavior

Blank page is opened in a new Window

### Actual Behavior

A new null reference exception on OnNavigatedTo ViewModel.Initialize because ViewModel is null.
It looks that the framework can not associate the Page DataContext with the configured ViewModel when the new window is created.

Download link to [ReproApp.zip](https://github.com/Microsoft/WindowsTemplateStudio/files/2314131/MultiViewCaliburnMicro.zip).

There is a similar problem with MVVMLight and multi-view #2487.

Can you have a look here? @nigel-sampson 
Thank you ;)

nigel-sampson:
Will take a look this weekend.

mvegaca:
Thank you very much :)

nigel-sampson:
I've tracked down the issue and a simple fix, just need to work out how to adapt in to the templates and check for some edges cases. Should have that for you soon.

What's happening is when using the `Frame` control and "view first" navigation it's the responsibly of the navigation service to instantiate the correct view model as the control navigates between views. Each navigation service wraps a specific `Frame`.

``` csharp
private async Task<ViewLifetimeControl> CreateViewLifetimeControlAsync(string windowTitle, Type pageType)
{
    ViewLifetimeControl viewControl = null;

    await CoreApplication.CreateNewView().Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () =>
    {
        viewControl = ViewLifetimeControl.CreateForCurrentView();
        viewControl.Title = windowTitle;
        viewControl.StartViewInUse();
        var frame = new Frame();
        var navService = new FrameAdapter(frame);
        frame.RequestedTheme = ThemeSelectorService.Theme;
        frame.Navigate(pageType, viewControl);
        Window.Current.Content = frame;
        Window.Current.Activate();
        ApplicationView.GetForCurrentView().Title = viewControl.Title;
    });

    return viewControl;
}
```

Simply adding the creation of the `navService` gives us the correct behavior.  We'll probably want to hold a reference to that service into `ViewLifetimeControl` though.

mvegaca:
Thank you @nigel-sampson I was trying that all works fine ;). The PR is approved and merged.

[...]

PR #2507, opened 2018-08-26
Resolve null reference for MultiView and Caliburn.Micro

nigel-sampson:
This resolves an issue raised in #2501 which is null reference when combining Caliburn.Micro and the MultiView feature.


nigel-sampson:
Having some issues running the tests locally, will resolve this soon.

Issue #2398, opened 2018-06-26
Review unit tests and code coverage


sibille:
Running minimum and rightclickWithLegacy tests code coverage for templates.core.dll is at 78%. Review what is not covered and do more tests.

sibille:
Moved to 3.0 as this is work is ongoing

crutkas:
Lets document which sections need to be done and create individual bugs here.  We're working on this but it dependent where we are currently working on the code.  individual issues will allow us to be more targeted.

once we have the issues created, xref and close this out

PR #2615, opened 2018-10-01
Issue2398 reviewunittests

sibille:
Quick summary of changes
- Added unit tests for postactions, ITemplateInfoExtensions
- Removed unused code

- #2398 Review unit tests and code coverage


mrlacey:
I'm curious about why there are files with different line endings.
As they've been added recently (based on the assumption that the affected tests used to work) is there a reason for them?

Could a mixture of different line endings cause potential issues elsewhere? (Wondering if there are potential consequences of just ignoring them.)

If we don't specifically need different line endings in different places, would it not be better to enforce that they're all consistent?

sibille:
This has been added due to tests failing on appveyor, we were doing this on other tests earlier too. I think once we're using azure piplines to build PRs we can remove this.

Issue #2640, opened 2018-10-10
Remove duplicate XElement.extensions file


GeertvanHorrik:
https://github.com/Microsoft/WindowsTemplateStudio/blob/8736d3c1c070a9e8274e6a034ac3f340c3749be6/code/src/Core/XElement.extensions.cs#L14

Code seems to have moved to /extensions/XElementExtensions.cs

GeertvanHorrik:
Nice, @crnd already created a PR for this one.

sibille:
Thanks @GeertvanHorrik  and @crnd!!

sibille:
Fixed in dev

Issue #2639, opened 2018-10-10
Remove duplicate Extensions.cs file


GeertvanHorrik:
https://github.com/Microsoft/WindowsTemplateStudio/blob/8736d3c1c070a9e8274e6a034ac3f340c3749be6/code/src/Core/Extensions.cs#L14

This code seems to be moved to /extensions/*, so can be removed?

GeertvanHorrik:
Included in PR.

PR #2648, opened 2018-10-11
Delete obsolete XElement.extensions.cs file

crnd:
The file has been previously copied to Extensions folder as
XElementExtensions.cs and the copied file is the one that is currently
being used. The file that is being deleted here is not included in any
of the csproj files anymore.

Related issue #2640


crnd:
There is also issue #2639 that has a similar file that is not included in any csproj files anymore. The contents of that file have also been moved to Extensions folder and split to several files.

PR #2723, opened 2018-10-31
Changes from executing manual tests after work done for 2.5

sibille:
Quick summary of changes:
- Fix links in docs and comments
- Sync vb and c# templates


mrlacey:
@sibille I'm looking at the WAD tests now. The issue is probably also due to scaling as dimensions are originally from a small screen. Have a solution for scaling that I'm testing.

Issue #2726, opened 2018-11-02
Warning message when building a project with a image gallery page


dgomezc:
When you build a project that includes a image gallery page you get a warning message: "XamlCompiler warning WMC1506: OneWay bindings require at least one of their steps to support raising notifications when their value changes"

sibille:
Verified on dev-nightly:
Templates version: 0.19.18311.1
Wizard version: 0.19.18310.2

PR #2728, opened 2018-11-02
Fix warning message to image gallery detail page

dgomezc:
## PR checklist

Quick summary of changes

#2726 - Warning message when building a project with a image gallery page


mrlacey:
Any reason this was set to OneWay originally?
It still seems to work with OneTime mode so wondering if I'm missing something.

sibille:
Not sure why it was added in the first place. We don't this binding to be oneway, because while images change and the collection is observable, a single image's properties won't change.

Issue #2774, opened 2018-11-27
WinUI:NavigationView does not navigate in Release Mode


Pinox:
Hi , something I noticed , the Hamburger menu items does not change the page views when app is build to UWP Release Mode.

 
![releasemode1](https://user-images.githubusercontent.com/5297255/49093489-dcd2ea80-f26c-11e8-8269-8bc1ac7168f0.PNG)

Using latest stable VS 15.9.2
![ts](https://user-images.githubusercontent.com/5297255/49093590-16a3f100-f26d-11e8-8dc7-8c7f6c4158c0.PNG)

crutkas:
@Pinox can you share the 'recipe' you used to aid us in debugging?

Pinox:
@crutkas    Sure,   Navigation Menu , Mvvmlight and then I had a couple of random pages as in screenshot above.

sibille:
@pinox, we're having a look at this. 
The issue seems to be related to using the EventTriggerBehavior on the ItemInvoked event. 

While we are working on a fix you can workaround by using the event instead. 
In ShellPage.xaml, remove the following code: 
```
<ic:EventTriggerBehavior EventName="ItemInvoked">
         <ic:InvokeCommandAction Command="{x:Bind ViewModel.ItemInvokedCommand}" />
 </ic:EventTriggerBehavior>
```
and add 

`ItemInvoked="NavigationView_ItemInvoked"`


In the ShellPage.xaml.cs add:

```
private void NavigationView_ItemInvoked(Microsoft.UI.Xaml.Controls.NavigationView sender, Microsoft.UI.Xaml.Controls.NavigationViewItemInvokedEventArgs args)
{
     ViewModel.ItemInvokedCommand.Execute(args);
}
```

Pinox:
@sibille Thanks.

crutkas:
Will this need a hot fix?

sibille:
Created an issue to investigate this at https://github.com/Microsoft/XamlBehaviors/issues/142. I think we should prepare a hotfix with the workaround while investigating the root cause.

mvegaca:
Hotfix added on a pull request
[https://github.com/Microsoft/WindowsTemplateStudio/pull/2776](https://github.com/Microsoft/WindowsTemplateStudio/pull/2776)

sibille:
Hotfix published

PedroLamas:
[...]

PR #2817, opened 2018-12-18
Add directions in Default.rd.xml

mvegaca:
WinUI:NavigationView does not navigate in Release Mode #2774


touseefbsb:
@sibille  @mvegaca  do we need to add directives for every event we put on NavigtionView? for example DisplayModeChanged and others? and what about nested controls, for example we have a button in footer of navigationview and we have command bind with that button ? ( apart from Loaded ofcourse )

mrlacey:
> @sibille @mvegaca do we need to add directives for every event we put on NavigtionView? for example DisplayModeChanged and others? and what about nested controls, for example we have a button in footer of navigationview and we have command bind with that button ? ( apart from Loaded ofcourse )

Only add them if your testing finds they're needed.

Issue #2794, opened 2018-12-07
Exception closing app using unmodified SuspendAndResumeService - Unable to remove the file to be replaced.


michael-hawker:
### Repro steps
Use [SuspendAndResume service](https://github.com/Microsoft/WindowsTemplateStudio/blob/dev/templates/Uwp/Features/SuspendAndResume/Services/SuspendAndResumeService.cs) from WTS as-is, no modifications or hooking into event.
Run app.
Close app.

(I've only seen this once after running app dozens of times.)

### Expected Behavior
App closes without exception.

### Actual Behavior

"Unable to remove the file to be replaced. (Exception from HRESULT: 0x80070497)"
>	MyProject.exe!MyProject.App.InitializeComponent.AnonymousMethod__16_0(object sender, Windows.UI.Xaml.UnhandledExceptionEventArgs e) Line 53	C#
 	[Native to Managed Transition]	
 	[Managed to Native Transition]	
 	System.Private.CoreLib.dll!System.Runtime.InteropServices.WindowsRuntime.WindowsRuntimeMarshal.RoReportUnhandledError(System.Runtime.InteropServices.WindowsRuntime.IRestrictedErrorInfo error)	Unknown
 	System.Private.CoreLib.dll!System.Runtime.InteropServices.WindowsRuntime.WindowsRuntimeMarshal.ReportUnhandledError(System.Exception e)	Unknown
 	System.Runtime.WindowsRuntime.dll!System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeCore()	Unknown
 	System.Runtime.WindowsRuntime.dll!System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeInContext(object thisObj)	Unknown
 	System.Private.CoreLib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state)	Unknown
 	System.Runtime.WindowsRuntime.dll!System.Threading.WinRTSynchronizationContextBase.Invoker.Invoke()	Unknown

StackTrace:
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #2836, opened 2019-01-09
Fix Suspend and Resume Service

dgomezc:
## PR checklist

Quick summary of changes

#2794 - Exception closing app using unmodified SuspendAndResumeService - Unable to remove the file to be replaced


crutkas:
@michael-hawker are you ok with this fix as is.  We are still investigating per the email thread for maybe a deeper fix

michael-hawker:
@crutkas yup, I think I added a loop in mine to try again to save, but this is a good starting point.  The only suggestion I have (assuming you don't mind changing the interface) is to have it return a `Task<bool>` so upstream could also handle the response.  

But don't know how you generally handle interface changes, so this is sufficient for a first step until we find the right folks in the platform to talk to about these calls.

crutkas:
@dgomezc can you do the change @michael-hawker  recommended?

dgomezc:
@crutkas @michael-hawker just added the changes.

Issue #924, opened 2017-08-10
Audit use of long namespaces


mrlacey:
There are places where classes are fully qualified with their namespace within the body of a class rather than being specified in a "using" clause at the top of the file.

Review and tidy up such occurrences.

mrlacey:
I don't think this affects generated code but need to check

mrlacey:
this is inconsistent with generated code too. Need to create a separate issue to tidy up that.

mrlacey:
Just reviewing generated code. Wizard code has been changed greatly since this was raised.

mrlacey:
also fixing some spelling errors in comments of generated code that were found during the manual review.

PR #2853, opened 2019-01-17
Removing explicit namespace use & fixing spellings in generated code

mrlacey:
For #924


mrlacey:
This includes template changes so can hold for 3.1 if it's too close to the 3.0 release.
Advise or set the milestone if you want to hold it.

sibille:
@mrlacey I think we can include this + thanks for the language review

Issue #2892, opened 2019-02-03
Template Studio is unable to create an app with whitespaces or hyphen in AppName


debjyotigorai:
I am getting the following error while creating an UWP app in C# using MVVW Basic. I think the error is related to "Setting Storage".
## Error
System.Exception: Error executing 'Microsoft.Templates.Core.PostActions.Catalog.Merge.MergePostAction'. Related template: wts.Feat.SettingsStorage. ---> System.IO.FileNotFoundException: There is no merge target for file 'C:\Users\Debjyoti Gorai\source\repos\DG DEV (UWP)\DG_DEV__UWP_\Strings\en-us\Resources_postaction.resw'. Related Template: 'wts.Feat.SettingsStorage'.
   at Microsoft.Templates.Core.PostActions.Catalog.Merge.MergePostAction.HandleFileNotFound(String originalFilePath, String suffix)
   at Microsoft.Templates.Core.PostActions.Catalog.Merge.MergePostAction.ExecuteInternal()
   at Microsoft.Templates.Core.PostActions.PostAction.Execute()
   --- End of inner exception stack trace ---
   at Microsoft.Templates.Core.PostActions.PostAction.Execute()
   at Microsoft.Templates.Core.Gen.GenController.ExecutePostActions(GenInfo genInfo, TemplateCreationResult generationResult)
   at Microsoft.Templates.Core.Gen.GenController.<GenerateItemsAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Templates.Core.Gen.NewProjectGenController.<UnsafeGenerateProjectAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Templates.UI.Services.GenerationService.<GenerateProjectAsync>d__5.MoveNext()

crutkas:
Possible to share the exact items you selected  outside of MVVM Basic?

I do agree it seems like a resource file for setting storage is missing.

[...]

Issue #1784, opened 2018-01-31
Some project names cause generated apps to not work


mrlacey:
Some project names cause problems when using the framework of the same name.

- Called **Prism** and uses the `Prism` framework causes the generated app to crash when started.
- Called **MVVMLight** and uses the `MVVMLight` framework causes NuGet restore to fail.
- Called **Caliburn.Micro** and uses the `Caliburn.Micro` framework causes NuGet restore to fail. (**Caliburn** as a name is fine.)

I forget and use these sometimes when making multiple similar versions of an app for debug/testing. Others may use them (for testing or otherwise) and it would be good to avoid them having issues.

Including these names in the `ReservedNamesValidator` should avoid the issue.

There are potentially lots of other project names that could be used that would cause errors but I don't think it's worth going looking for them. We can add any more when encountered but there are better ways to spend time right now.

crutkas:
Good bug!

ralarcon:
Agree, @mrlacey!

mrlacey:
I thought this was going to be a simple fix, but it isn't.

The wizard only does validation on item names, not on the project name (which becomes the default namespace.) The project name is selected before the wizard starts.

I see four options available

Option 1. Modify the project name so it doesn't have issues.     
This could be done by appending a value to the name (e.g. "Prism" would become "Prism1" or "Prism_", etc.) The downside here is we'd be changing the value they want but without explaining why and they may not notice.

Option 2. Try and workaround the issue without changing the project name.   
This could be done by altering the code generation to append something to the name where it's used as part of a namespace.  
The downsides to this are that: we may end up doing something non-obvious with the generated code; it would make right-click>Add support more complicated (though needing to determine what, if anything should be done to make the root namespace different from the project name); it would also, I think, require some modifications to the current templates as some use the `Param_RootNamespace` and `Param_ItemNamespace` interchangeably but this would not always be possible in this scenario. This would also add further potential complication to template authoring.

Option 3. Prevent the generation of a project that has a potentially problematic name.  
Display an error message and prevent the wizard from creating an app.   
An error message something like "The project name you have chosen is likely to cause namespace collisions or other issues with creating a working project. Please chose a different name and try again."

Option 4. Like option 3 but just give a warning, and any issues are up to them to resolve.  
This feels like a cop-out. I also think it is likely to create more support issues where something gets generated, doesn't work correctly, and then we spend ages trying to track down the problem only for them to claim they didn't see the message.



[...]

Issue #299, opened 2017-05-11
Add unit test project to solution


mrlacey:
Add a feature to add a unit test project to the solution. This should include a reference to the app project.

Need to consider what, if any, test(s) to include by default.

jamesmcroft:
Similar to my comments in #300 - would this be optional as it currently is when creating a new blank project?

crutkas:
why not just allow someone to opt in as a system level and make the whole project straight up testable.   With that said, I can see how to do UI testing, but to do stub out a proper unit test without any business logic?  feels odd as not sure how much would be testable

mrlacey:
My thinking was that by making this a "feature" it becomes optional.
Yes, there's not a lot we can test but by making it easy to add unit testing to the solution we can hopefully help with its adoption.

Even if we only added a single test which included `Assert.Inconclusive("TODO: Write some tests");` I would consider this better than doing nothing.

magol:
Perhaps add some commented examples of some types of tests that you can perform just to show how to best write tests based on selected conditions.

The conditions for writing tests differ depending on the chosen framework (Code behind, MVVM Basic, MVVM light, PRISM), how Dependency injection is performed,  chosen unit test and mock framework, etc.

mrlacey:
Once we get to FCU as a minimum SDK and split out potentially X-Plat functionality this could be 3 options:

- [ ] Add a UWP unit test project that references the main app for testing the functionality it contains
- [ ] A NetCore 2.0 project using MSTest that references the X-Plat library for testing X-Plat functionality
- [ ] A NetCore 2.0 project using XUnit that references the X-Plat library for testing X-Plat functionality

mrlacey:
moving back to backlog as timeline for major restructuring is now undetermined

crutkas:
@mrlacey, will this be possible to be done for 3.0?

mrlacey:
Thinking about what to name the projects added to the solution and need them to be unique (so we can add them all to a solution--even if only in testing) but meaningful.

I'm thinking:
- `{App}.Tests.Unit.MSTest` (would reference `{App}.Core`)
- `{App}.Tests.Unit.NUnit` (would reference `{App}.Core`)
- `{App}.Tests.Unit.xUnit` (would reference `{App}.Core`)
- `{App}.Tests` (Would reference `{App}` and be a "Universal Windows Unit Test App")

[...]

PR #2894, opened 2019-02-04
Fix bug preventing app generation when name includes hyphen, space, etc.

mrlacey:
For #2892

Avoids using project names directly in a namespace without first making "safe"

Makes existing templates consistent.
Ensures converting project name to namespace follows standard VS conventions.
Added tests to verify the above and can generate apps with project names that include non-alphanumeric characters.

Some of this had already been done for #299 

X-Ref #1784 which covers updating the wizard to warn about project names that can cause problems.

The ability to support project names that included a hyphen (and possibly other non-alphanumeric characters) apparently worked before 3.0 was released. I can't confirm this but if true, there is a case for this going in a **HOTFIX**.

### DIFF
--- a/code/src/Core/Extensions/StringExtensions.cs
+++ b/code/src/Core/Extensions/StringExtensions.cs
@@ -51,5 +51,25 @@ namespace Microsoft.Templates.Core
         {
             return value.GetMultiValue().Length > 1;
         }
+
+        // This is the same substitution as VS makes with new projects
+        public static string MakeSafeForProjectName(this string value)
+        {
+            var result = new StringBuilder();
+
+            foreach (var character in value)
+            {
+                if (char.IsLetterOrDigit(character) || character == '.')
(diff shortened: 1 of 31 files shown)


sibille:
@mrlacey, this solves the issue for the emulator and tests, where we did not make the projectname safe before using it on the namespaces. 
In the extension we're having just the oposite problem. We're using the $safeprojectname$ from VS replacements dictionary, which is fine for the namespaces, but not for the folder names. 
I think we'll have to use both, $projectname for path and project name, $safeprojectname$ for the namespaces. In the extension we get both values from VisualStudio, in the emulator we can use your extension method.
The underlying issue has been around for a time, but with the multiproject this is causing the extension to fail, as we generate in different directories. I think we should ship a hotfix for this.

crutkas:
lets spin up the hot fix

PR #2911, opened 2019-02-07
Create client-authentication-msal-example.md

ClairelyClaire:
Spec document for UWP authentication client using Microsoft Authentication Library (MSAL) and AzureAD.

## PR checklist

Quick summary of changes

- Have automated tests been added for new features?
- If you've changed the UI: 
- If you've included a new template: 
  - Be sure you reviewed the [Template Verification Checklist](https://github.com/Microsoft/WindowsTemplateStudio/wiki/Template-Verification-Checklist).
- Have you raised issues for any needed follow-on work?
- Have docs been updated?
- If breaking changes or different ways of doing things have been introduced, have they been communicated widely?

### DIFF
--- a/docs/features/client-authentication-msal-example.md
+++ b/docs/features/client-authentication-msal-example.md
@@ -1,0 +1,564 @@
+Introduction
+============
+
+Azure AD Authentication – UWP C\# Code Spec
+-------------------------------------------
+
+This document details the minimum generated C\# code required to implement Azure
+AD (AAD) auth in a UWP C\# app.
+
+This spec includes a couple of optional components – integrated authentication,
+and an example function that queries Microsoft Graph and retrieves the logged-in
+user’s basic profile information. This isn’t necessary for an app to implement
+AAD authentication, but it’s useful to developers, in that it provides an easy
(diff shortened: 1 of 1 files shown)


crutkas:
This is our spec, the others in the "feature" folder are true items we link to, correct?  I think this needs to be moved into a "spec" folder then

I know the title says "spec" but want to make reduce any confusion.

crutkas:
* can be as simple as features/spec for the folder

sibille:
oh, .. I did not check where it went to. The other spec document is in a specs folder on the root https://github.com/Microsoft/WindowsTemplateStudio/tree/dev/specs
Is that ok? Should it be specs/features?

mrlacey:
Why are specs getting their own checked in files?
Why not just track these things in issues?
There needs to be issues for spec'd items anyway and it would put specs and comments in the same place.

crutkas:
We want to start being more transparent.  I think if we use the docs/feature folder for production features, i don't want to confuse the two.  I think having a 6 page spec in the issue itself gets overwhelming along with tracking changes for the spec.

We're trying to see what works / doesn't work.

Issue #2994, opened 2019-03-07
Composition query $feature and $page always return false when no features/pages added


sibille:
Composition query containing $feature != "wts.Feature.SomeFeature" returns false when no features are added, this should return true

sibille:
Verified in dev-nightly:
Templates version: 0.20.19073.1
Wizard version: 0.20.19073.1

PR #2995, opened 2019-03-07
Fix compare method in QueryableProperty

dgomezc:
## PR checklist

Quick summary of changes

#2994 - Composition query $feature and $page always return false when no features/pages added


mrlacey:
Feels like this shoudl have includedd some new or updated unit tests related to this functionality,

sibille:
Agree, we'll add them in a new PR asap

Issue #3023, opened 2019-03-19
Add UI Testing for generated MenuBar apps


mrlacey:
Add language & framework `VisualComparisonTests`.

X-Ref #2292 & #2999

mrlacey:
Adding Navigation tests now to help with checking changes made for #3075

mrlacey:
closing as merged to dev and can now be used in other feature dev

PR #3108, opened 2019-04-24
Add UI tests for apps using MenuBar navigation

mrlacey:
For #3023 

Ensures that all CS frameworks produce the same UI for each page.
Ensures that VB & C# apps using the same framework produce identical UIs for each page.

Does not include Login pages for Identity features.


mrlacey:
This just contains tests so can go in 3.2.

Issue #3075, opened 2019-04-02
Load all data asynchronously, not in constructors


mrlacey:
As discussed in https://github.com/Microsoft/WindowsTemplateStudio/issues/900#issuecomment-478519871

Many pages currently load data in page constructors. This is fine for the sample data but doesn't work with real-world scenarios where data comes from the disk or across a network.
In those scenarios, the loading of the data may take time and so should be done asynchronously.

This task is to change all pages that load data so that they do so asynchronously and can, therefore, more easily be used with data from "real" sources (such as are being added in #900 & #133)

To Change:
- [x] Chart page
- [x] DataGrid page
- [x] Telerik Grid page
- [x] Image Gallery Page
- [x] Content Grid Page
- [x] ~MasterDetail Page~
- [x] Core Library MSTest project feature
- [x] Core Library NUnit project feature
- [x] Core Library xUnit project feature

Applies to all frameworks in C# & VB

MasterDetail pages already load their data asynchronously.

mrlacey:
in final testing

mrlacey:
PR raised and work for #900 being done on top of this branch

mrlacey:
verified in 
Templates version: 0.20.19140.3
Wizard version: 0.20.19138.1

PR #3116, opened 2019-04-29
Pages now get from SampleDataService using async methods

mrlacey:
For #3075 

C# & VB.Net (All frameworks)
All pages and test projects:
- Chart page
- DataGrid page
- Telerik Grid page
- Image Gallery Page
- Content Grid Page
- Core Library MSTest project feature
- Core Library NUnit project feature
- Core Library xUnit project feature

MasterDetail Page was already loading data in an async way

Changes have been extensively tested manually and with automated tests (also updated as part of these changes)


mrlacey:
Replaced with #3123 which also includes these changes.

Issue #3181, opened 2019-06-13
Mobile App Service Enhancements


sibille:
We had a look at the Mobile App Service and have the following questions/proposals: 
- this should have more comments on how to use, developer TODOs + links to web api  docs
- we should change [Route("api/item")] to [Route("api/[controller]")] as Visual Studio does when creating a web api 
- Can we change the name to "Web API" instead of "Mobile App Service"?
- Should we have a doc on how things play together in different scenarios?

  - UWP -> HttpDataService -> API Controller ->  SampleDataService
  - UWP -> HttpDataService -> API Controller -> SqlDataService
  - UWP -> SQLDataService

sibille:
Deleted first point as this is the same as #3180

sibille:
Reviewing the nuget package updates we came up with a few questions more: 
- Should this be aspnet core 2.2?
- Should use **Metapackage Microsoft.AspNetCore.App** instead of  Microsoft.AspNetCore.App Version="2.1.1" and Microsoft.EntityFrameworkCore Version="2.1.11"

mrlacey:
> * Should this be aspnet core 2.2?

Targetting 2.1 (rather than 2.2) as 2.1 is the Long Term Servicing (LTS) version. Want to push people to the version that is supported for longer by default. They can upgrade if they wish.

mrlacey:
verified in 
Templates version: 0.20.19186.2
Wizard version: 0.20.19186.2

Issue #3180, opened 2019-06-13
Web API Feature improvement


mrlacey:
Following on from #133

Change the API project to use the same `SampleData` items as all other pages and features that use sample data.

mrlacey:
verified in 
Templates version: 0.20.19185.1
Wizard version: 0.20.19185.1

PR #3227, opened 2019-07-02
Web API  Improvements

mrlacey:
For #3180 & #3181

- Renamed from 'Mobile App Service'
- Now using SampleDataService & SampleOrder
- Removed EntityFrameworkCore reference
- Restructured templates to allow for Prism variations
- Added `TODO WTS` comments
- Updated route definition


sibille:
@mrlacey, can we use the Metapackage Microsoft.AspNetCore.App? I don't think the postaction supports metapackages, but think we can add it on the csproj file directly as vs does when you create a asp.net core web api.

<PackageReference Include="Microsoft.AspNetCore.App" />
(https://docs.microsoft.com/en-us/aspnet/core/fundamentals/metapackage-app?view=aspnetcore-2.1)

mrlacey:
> @mrlacey, can we use the Metapackage Microsoft.AspNetCore.App? I don't think the postaction supports metapackages, but think we can add it on the csproj file directly as vs does when you create a asp.net core web api.
> 
> (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/metapackage-app?view=aspnetcore-2.1)

This is included.
screenshot from app I just generated.
image

Do you mean to update to the 2.2.5 version?

sibille:
No, I'm saying have `<PackageReference Include="Microsoft.AspNetCore.App" />` on the csproj instead of `<PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.11"/>`.

templates/Uwp/Services/SqlServerDataService.Prism/.template.config/template.json
```diff
@@ -14,7 +14,7 @@
     "wts.platform": "Uwp",
     "wts.group": "Data",
     "wts.version": "1.0.0",
-    "wts.displayOrder": "1",
+    "wts.displayOrder": "s",
```
sibille: Should this be a 2?

mrlacey: Yes. Will also update the validator tests as they should have picked that up.

mrlacey: There is a test that should have caught this. Now to investigate why it didn't...

sibille:
Bringing in those changes, we can adapt the package reference later

PR #3230, opened 2019-07-03
Further Web API improvements

mrlacey:
For #3181 
(and a continuation from #3227)

- Adds a metapackage reference (no version number) to `Microsoft.AspNetCore.App`
- also fixed automated testing of WebApi project with StyleCop

**Note**
I'm having issues with SSL in a dev environment on my machine and so can't test this fully at the moment.
Equivalent Web API project created as part of a Xamarin app also has the same error, so hoping this is something on my machine and not in the templates. I'm not aware of having changed anything that might impact this so hoping it's not an issue with the templates.

### DIFF
--- a/_utils/List-CSharp-Templates-Without-VisualBasic-Equivalents.ps1
+++ b/_utils/List-CSharp-Templates-Without-VisualBasic-Equivalents.ps1
@@ -2,7 +2,7 @@
 # Any output indicates missing VB templates.
 
 # Get list of all templates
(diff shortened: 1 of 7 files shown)


sibille:
@mrlacey, its failing for me too, in this pr and in dev. 
I think its related to the launchsettings.json.
MobileAppService Configuration is working fine (can we change the name to WebApi here too?), but IIS Express Configuration is failing.

Underlying reason seems to be the SSL port, which should be starting with 443*, as IIS Express needs a port in the range 44300 to 44399. (https://docs.microsoft.com/en-us/iis/extensions/using-iis-express/handling-url-binding-failures-in-iis-express)

Issue #1246, opened 2017-10-06
Treeview based on master/detail Page


crutkas:
A user made a great request.  #1237

We have navview, tabbed, and blank.  A treeview is a common pattern and maybe we should think / come up with a POC of what this may look / act like

mrlacey:
Why a nav option and not a page?
Seems more like a variation on master/detail but with a treeview.
Of course, we may want to wait for the official treeview control to finally ship before doing this.......

crutkas:
I agree actually after thinking about this for a bit.  It feels much more master detail based

mvegaca:
This is a great idea but a treeview need to define navigation items in navigation items groups and to add we will need to think about how user specify those groups and decide what page must been on each group. We could also add only one group and generate all pages on the same group. Or create one group for each page type: Maps, Galleries, etc.

mrlacey:
My comparison to Master/Detail was more to highlight that it feels more appropriate to put this as a page rather than a project type.
"MasterDetail" is a perfectly valid navigation pattern and project type. We don't make this a project type though. However, it would be easy for an app to be developed with WTS that does this. Just create a blank app with a single MasterDetails page.

I am not proposing that any page based on a TreeView control (when it finally ships) should directly mimic the "Master" part of the MasterDetails page/functionality.
If we started building navigation functionality into a page with a TreeView control on it that would be being too specialized.

Let's wait until a final version of the control is released and then we can discuss what is best to include on a page with one.

mvegaca:
Agree

mrlacey:
Needs a new data type in sample data - TBD.
Assuming this control will be in WinUI library to give the version level support we want for including this.

mvegaca:
TreeView Proof of concept is ready for review on [this branch](https://github.com/mvegaca/TreeViewPoC/tree/dev). This is a first approach and we would like to get some feedback from WTS users.

We have added more detailed SampleData to show a hierarchical model (Company / Orders / OrderDetails). These data enhancements will also improve other pages that use sample data like MasterDetial or Data Grid pages.

mvegaca:
TreeView page short and long description proposals:

**Short description (shown in card)**

The tree-view page has a tree view control to show hierarchical content.

**Long description (shown in details page)**

The tree-view page has a tree view control to show hierarchical content. When an item in the tree view is selected, it's details are shown in a details pane. This pattern is frequently used to show hierarchical relationship between items.

[...]

PR #3249, opened 2019-07-05
Update pages order

mvegaca:
Treeview based on master/detail Page #1246


mrlacey:
What's the new page order?

There are lots of changes here that are just white-space and formatting changes. I presume these were made by the editor used while making other changes. What can we put in place to ensure consistency of white-space in template files so that any future changes don't also end up including lots of potentially distracting other changes?

sibille:
We added the TreeView page after the MasterDetail page.
Regarding the whitespaces, they are removed manually, with help of the Trailing Spaces Extension of VSCode.
Standardizing json format seems a good option. Will have a look at tools for this, any suggestions?

mrlacey:
> Will have a look at tools for this, any suggestions?

My 1st thought is to use [editorconfig](https://github.com/editorconfig/editorconfig-vscode) but not sure it has properties for everything we need.

Issue #3296, opened 2019-08-09
Remove <AppDesignerFolder> in the default vbproj template


Nukepayload2:
The default app designer folder of a VB project is "My Project", not "Properties". So, I suggest removing [this line of code](https://github.com/microsoft/WindowsTemplateStudio/blob/e9bab99691c7949bb7942d7c7dbbdef2769fb26d/templates/Uwp/Projects/Default._VB/wts.DefaultProject.vbproj#L10) .

mrlacey:
It doesn't hurt having it there (AFAIK) but risks confusion so will take it out.

PR #3297, opened 2019-08-09
Remove AppDesignerFolder from proj file

mrlacey:
For #3296

Removes project setting that applies to C# and not VB.

No other/related changes necessary.


mrlacey:
Forcing the PR build (again) as can't see why anything in this change should cause tests to take longer.

mrlacey:
Not sure why the CI build is failing at 2 hours.
Even sequentially, I can run all the same tests in ~56 minutes.
Will not try and force again (it's had 3 attempts now)
Hoping someone with access to more detailed server logs can identify an issue.

sibille:
One of our build machines has performance issues, that's why the build timed out. I disabled the "slow" agents until I can reset the build machine. Queued a new manual build at https://winappstudio.visualstudio.com/WTS/_build/results?buildId=30973 which ran fine.

PR #3303, opened 2019-08-31
Documentation updates

mrlacey:
## PR checklist

This is the result of a thorough review of documentation. It hasn't made everything perfect but it is hopefully a positive step in the right direction.
Documentation is something that will always have the opportunity for improvements and I expect further improvements in the future. There are issues relating to documentation that are still on the backlog but they are for specific issues/requests that can be actioned at a later time.

This PR includes changes relating to:

- review and update all files under /docs
- Every item in the wizard now has a page on /docs
- grouping docs pages and links at the bottom of pages
- Standardized white space on all pages
- change all links to docs.microsoft.com to be locale agnostic
- standardize on using `vb` (rather than `vbnet`) in markdown files
- add alt text for all images
- standardize on using `WinTS` as the project abbreviation (rather than `WTS`)

Quick summary of changes

- Which issue does this PR relate to? #2294 #3284 #3285 #3286 #3287
- Anything that requires particular review or attention? **no**
- Do all automated tests pass? **yes**
- Have automated tests been added for new features? **n/a**
- If you've changed the UI: **n/a**
- If you've included a new template: **n/a**
- Have you raised issues for any needed follow-on work? **n/a**
- Have docs been updated? **YES**
- If breaking changes or different ways of doing things have been introduced, have they been communicated widely? **n/a**

Issue #1997, opened 2018-02-27
Build dev.templates.tests.wack_20180227.1 failed


wasteam:
## Build dev.templates.tests.wack_20180227.1
- **Build result:** `failed`
- **Build queued:** 2/27/2018 12:59:03 PM
- **Build duration:** 1.65 minutes
### Details
Build [dev.templates.tests.wack_20180227.1](https://winappstudio.visualstudio.com/web/build.aspx?pcguid=a4ef43be-68ce-4195-a619-079b4d9834c2&builduri=vstfs%3a%2f%2f%2fBuild%2fBuild%2f25149) failed

+ Microsoft.Common.CurrentVersion.targets (1124): C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(1124,5): Error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.7" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend.
+ Microsoft.Common.CurrentVersion.targets (1124): C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(1124,5): Error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.7" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend.
[...]

PR #3304, opened 2019-08-31
Issue1997 multi instance

mrlacey:
For #1997

Add features for launching multiple instances of the app. There's a basic version which always forces a new version, and and advanced version that allows control over whether an existing instance is reused or a new instance of the app is created.

This was previously blocked by a dependency on having RS4 as the minimum target platform in the generated app.

This is currently a draft as it will require modifications to associated docs once #3303 is merged.


mrlacey:
This is ready for review but is likely to require some manual intervention to merge with the changes made for #3305 (if that is merged first)

sibille:
@mrlacey, should this be merged first or the #3305

mrlacey:
> @mrlacey, should this be merged first or the #3305?

either - I'll fix up whichever isn't 1st

Issue #3259, opened 2019-07-08
Update  TargetPlatformVersion to 18362


sibille:
Update TargetPlatformVersion to 10.0.18362.0 and set TargetPlatformMinVersion Version to 10.0.17763.0

sibille:
Verified in dev-nightly; 
Templates version: 0.20.19263.2
Wizard version: 0.20.19262.1

PR #3308, opened 2019-09-04
Set target version to 10.0.18362.0

mvegaca:
Update TargetPlatformVersion to 18362 #3259


mrlacey:
This increases just the target version and not the min target version too

Will that change separately or is this a deliberate change to start supporting the latest release and the two prior to that (n&n-2)?

Changing to 'n&n-2' will have an impact on future support for new platform functionality.

There are probably also documents that need updating that relate to this change too.

sibille:
Sorry, I was a bit too short on the issue. We should update both, Target and MinPlatform Version @mvegaca

mvegaca:
Done!

Issue #3228, opened 2019-07-02
Improve testing of generated templates


mrlacey:
The current testing of templates is a long and slow process. There are a few scripts to enable running parallelizing of some test execution but we can (and should) do better.

Need to do more to fully scope this out but the goals are:
- make tests faster to run (to avoid excuses for not running them)
- make tests easier to run.
- avoid developers relying on automated test execution at PR (or post PR) time
- Make it easier to test small subsets of templates. This may require reviewing data-bound test generation (which is slow anyway)

mrlacey:
@sibille Please can you verify that the details in https://github.com/microsoft/WindowsTemplateStudio/blob/dev/docs/getting-started-developers.md#test-execution are correct in terms of which `ExecutionSet`s are run where/when?

I want to make sure that any changes made for this issue don't break any of the automated test execution. Thanks.

sibille:
@mrlacey, just checked, docs are up to date.

mrlacey:
Adding "meta" executionsets.
These are named matching the [groupings executed during the different pipeline operations](https://github.com/microsoft/WindowsTemplateStudio/blob/dev/docs/getting-started-developers.md#test-execution) but beginning with an underscore.

The aim is to make it easier in VS to verify all tests that will be executed on the server will pass.

![image](https://user-images.githubusercontent.com/189547/63769566-09cf2f00-c8cb-11e9-9bcb-d0021771c534.png)

mrlacey:
Rather than list template IDs that do support both C# & VB, just list the ones that don't support VB.
This removes the need to add to the list every time a template is added which should mean this isn't missed/forgotten and testing is more likely to catch differences between languages

mrlacey:
To improve the speed of the CI builds, I propose moving the `GenerateAllPagesAndFeaturesAndCheckWithStyleCopAsyncWithOptionalLogin` tests to just be part of the full test suite.

Thoughts?

sibille:
I added this to the CI build to cover all templates with styles there, but am fine with moving it to the full test suite.

mrlacey:
>> To improve the speed of the CI builds, I propose moving the `GenerateAllPagesAndFeaturesAndCheckWithStyleCopAsyncWithOptionalLogin` tests to just be part of the full test suite.
>> 
>> Thoughts?

> I added this to the CI build to cover all templates with styles there, but am fine with moving it to the full test suite.

[...]

PR #3340, opened 2019-10-01
Testing changes

mrlacey:
For #3228

This contains:
- misc tidying of tests
- adding new execution sets to see what is run when
- Change language comparison tests to exclude templates that aren't in C# & VB rather than listing those that are.
- Ensure appropriate language is used when finding tests. (Some helper methods had hardcoded C# language lookup so tests that were supposed to verify VB behavior when actually looking at C# templates)
- Fix assorted issues found when running tests.

Just a draft as it will need updating once the VB TabView templates are merged


mrlacey:
This includes fixes for the following templates:
- TabView
- TreeView
- UserActivity
- HttpDataService

PR #3349, opened 2019-10-07
Missing nuget package on TabView VB Templates

sibille:


mrlacey:
This looks like a duplication of functionality that was added in https://github.com/microsoft/WindowsTemplateStudio/commit/9178b9e85050c401c2332fa569c56bb7ddd3ea2d

Is this really needed?

sibille:
The nuget packages are missing. Test Name:	BuildCodeBehindOneByOneItemsVBAsync(itemName: "TabView", projectType: "Blank", framework: "CodeBehind", platform: "Uwp", itemId: "wts.Page.TabView.CodeBehind.VB", language: "VisualBasic") detected that.

mrlacey:
The commit I referenced fixes another issue that was failing for both C# & VB versions in a different test.
Don't know why that test passed previously for me though. - Must have been before I found that separate issue where VB tests where actually using C# templates. (The fix for that was merged after the one for TabView.VB so could explain the previous test passes.)
I think that starting from a partial VB template implementation also made creating the VB versions more complicated. In future I recommend not creating empty files just to have some tests pass. Where there are tests that check for versions of templates in C# & VB exclusions are possible for things that only exist in C#. It is not a good practice to do enough to satisfy some trivial tests when the functionality hasn't been implemented and other tests will fail.

Hopefully only my time has been wasted on this.

sibille:
Agree, we'll not add again empty files just to satisfy a test as this seems to be the underlying reason.

Issue #3347, opened 2019-10-07
Combine template tests 


sibille:
With the growing permutation of things to test we should combine some tests and test as much as we can on each generated project. 
The following tests can be combined:
- `BuildEmptyProjectAndInferConfigAsync` and `BuildEmptyProjectAsync`
- `BuildAllPagesAndFeaturesAsync` and `BuildAllPagesAndFeaturesThenRunTestsAsync` with the `GenerateAllPagesAndFeaturesAndCheckWithStyleCopAsync` and `GenerateAllPagesAndFeaturesForcedLoginAndCheckWithVBStyleAsync` tests
- `BuildAllPagesAndFeaturesProjectNameValidationAsync` and `BuildAllPagesAndFeaturesRandomNamesAsync`

mrlacey:
this might also be a good opportunity to also tidy up / address the warnings in the test output as this might help make reading test output easier.

mrlacey:
Combining `BuildAllPagesAndFeaturesProjectNameValidationAsync` and `BuildAllPagesAndFeaturesRandomNamesAsync` means that `BuildAllPagesAndFeaturesRandomNamesAsync` will no longer be part of the CI tests.
Including the combined test as part of the CI set would increase CI test execution time and part of the reason for making these changes is to reduce execution time.

sibille:
I think that's ok, stylecop tests should find most of the issues on pr builds. 
Full tests are triggered by each integration to dev/master, so if sth breaks we'll catch and fix.

mrlacey:
closing as merged and have been executed in the nightly CI/DevOps process without issue

PR #3383, opened 2019-10-22
Combine template tests

mrlacey:
For #3347 

Combines some of the tests which have a large amount of overlap. While separate tests is better in a number of scenarios, these tests are hardly ever run separately and so the benefit of being able to have narrowly scoped tests loses out to having a test suite that takes less time to run.

Refactoring tests also found a bug where OptionalLogin on Prism didn't cope with the main page being renamed. There is a fix for this as part of this PR.

The old test should have caught this before but at least we know this is passing now.

Once this is merged, will need a full (overnight) build to verify everything still behaves as before.


mrlacey:
This is scheduled for 3.6 but submitting now because there is a benefit as soon as this is merged.

Issue #3345, opened 2019-10-03
Improve WACK testing


mrlacey:
Follow on from #3228 

These tests are currently a bit tempermental and fragile. Improve this.

sibille:
@mrlacey, can we speed up the wack tests somehow?
Ideas: 
- Exclude the test projects 
- Only generate the x86 msixbundle

If not, can we split them in two ExecutionSets to do separate wack tests with BackgroundTask and MultiInstance?

mrlacey:
All merged in and seem to be working ok

Issue #2574, opened 2018-09-19
Re-include background task to WACK


sibille:
We'll exclude background tasks from WACK tests due to issue #2358. Opened this issue to re-include once it's fixed.

mrlacey:
#2358 has been closed but I still get the incorrect WACK failure on a machine with 1803 (the original issue).
Reenabling this will be dependent upon what is running on the build machine

sibille:
Tested WACKs with BackgroundTask and it's running fine, but BackgroundTask and MultiView cannot be tested at once. Moving this to 3.6 to see if we can speed up or separate Wack tests somehow (while looking at https://github.com/microsoft/WindowsTemplateStudio/issues/3345) and test both scenarios.

sibille:
Verified in dev

PR #3403, opened 2019-10-29
Improve WACK tests

mrlacey:
For #3345 

- Reduce overall time taken to do WACK tests
- Ensure coverage for all items that may be affected by WACK testing


mrlacey:
These changes also cover #2574

Issue #3437, opened 2019-11-18
Dev-nightly: Generation fails when adding all features to a Prism app


sibille:
1. Create a new app with framework Prism
2. Add all features

-> Generation fails as line "await Task.CompletedTask;" could not be found

sibille:
Verified on dev-nightly: 
Templates version: 0.20.19331.1
Wizard version: 0.20.19331.1

PR #3439, opened 2019-11-18
Mark line await Task.CompletedTask; as optional context

sibille:
Quick summary of changes
- Mark line "await Task.CompletedTask;" as optional context

#3437 Dev-nightly: Generation fails when adding all features on a Prism app

### DIFF
--- a/templates/Uwp/Features/BackgroundTask.Prism/Param_ProjectName/App_postaction.xaml.cs
+++ b/templates/Uwp/Features/BackgroundTask.Prism/Param_ProjectName/App_postaction.xaml.cs
@@ -23,7 +23,9 @@ namespace Param_RootNamespace
 
         protected override async Task OnActivateApplicationAsync(IActivatedEventArgs args)
         {
+//{??{
             await Task.CompletedTask;
+//}??}
         }
 
 //{[{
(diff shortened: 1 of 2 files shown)


mrlacey:
Was this found with automated testing?
If so, which one?
If not, could it have been?

sibille:
No, was found in a manual test. It only happens if you add first a feature that removes the  
"await Task.CompletedTask;" from the OnActivateApplicationAsync method, and then either add BackgroundTask or CommandLine. Tests that add all features work as they add features in alphabetic order. We cannot cover all variantions though. 
What we could do is check the templates, if a template removes any line, it should only be used as optional context in every other template of the same framework + projecttype.

mrlacey:
> No, this was found in a manual test. It only happens if you add first a feature that removes the
> "await Task.CompletedTask;" from the OnActivateApplicationAsync method, and then either add BackgroundTask or CommandLine. Tests that add all features work as they add features in alphabetic order. We cannot cover all variantions though.
> What we could do is check the templates, if a template removes any line, it should only be used as optional context in every other template of the same framework + projecttype.

Explains why I couldn't repro the issue, and why the automated tests found nothing.

Something that looks at conventions in the templates & expectations about similar templates may be appropriate.

PR #3465, opened 2019-12-04
Fix release build

sibille:


mrlacey:
Realizing this is old, but for future reference or if this comes up again.

As these using directives are only used in one place, inside the Release build. wouldn't it be better to include the namespaces in the usage and not need the directives? This would make it clear why they show up as unused under the debug (default--for developers) configuration.

Issue #3068, opened 2019-04-01
will prism support be dropped from wts?


touseefbsb:
Just found out that prism is officially dropping support for uwp platform 
( source : https://github.com/PrismLibrary/Prism/issues/1745 )
So the question is "will windows template studio follow the same path?"

mrlacey:
We will work with the Prism team to work out the best way forward.

mrlacey:
Nothing yet on [the Template10 site](https://github.com/Windows-XAML/Template10#about-template-102).

SimoneBWS:
I love and use a lot Prism for WPF. Now I'm starting to write a new UWP app (a sort of porting from WPF to UWP), which MVVM framework do you recommend to use?

mrlacey:
> I love and use a lot Prism for WPF. Now I'm starting to write a new UWP app (a sort of porting from WPF to UWP), which MVVM framework do you recommend to use?

Windows Template Studio does not recommend one framework over another. Multiple options are provided so that people can choose the one that is right for them.  
Admittedly, that doesn't help your situation but the answer there is probably "it depends". With the uncertainty over the future support of UWP from Prism (or a direct equivalent) you may want to look at alternatives. The individual features and functionality of Prism that you currently use will play a part in determining which alternatives are most appropriate. (This is part of the reason there's no simple response.)


Perhaps other Prism users who have already made this decision can share their thoughts. Alternatively (or additionally) you may want to also ask this question ("I used Prism for WPF but now want to port to UWP, what should I use?") of the [Prism team](https://github.com/PrismLibrary/Prism/) themselves.

SimoneBWS:
Thanks for your reply @mrlacey. I'll ask a suggestion to the team, but I hope that Prism for UWP remains supported and mostly aligned with others platforms versions (WPF and Xamarin).

I need more and more to bring large LOB apps (or at least some parts of them) from WPF to UWP, and Prism for WPF is great (in my opinion)!

christiannagel:
Updating a project generated with Prism to the newest stable version of Prism.Unity (7.2.0.1367) breaks the build. The 7. versions of Prism have dependencies to .NET Core 3 or .NET Framework 4.5, the dependency for UAP v10.0 has been dropped - this is only available in the March-2017 released version 6.3.
I think Prism should be dropped from Windows Template Studio, this is not a good experience.

OzBob:
[...]

PR #3496, opened 2019-12-30
Add note on Prism/UWP support

mrlacey:
For #3068

Adds a note that Prism have stopped supporting UWP.

Just a docs change. No code or tests are impacted.


sibille:
@mrlacey: Thanks for adding this note. Can we tweak this a bit and just say: 

> **Note.** Prism does not support for UWP with their version 7 release. Windows Template Studio continues to use version 6.3.

Version 6.3 is valid for UWP and I don't want to speculate regarding the future.

PR #3599, opened 2020-03-21
Replaced MSTest asset files with link to main project.

mklemarczyk:
**Quick summary of changes**
MS Test project should refer to main project assets files.
The real files was replaced by project link in this PR.

The change should be tested if the links are resolved properly.


mrlacey:
> MS Test project should refer to main project assets files.

Why _should_ they?

If the assets in the main app are replaced, deleted, or renamed this could break the linked files in the test project.
With one of the goals of WinTS being to make it easy for people to get started. I have a concern this makes it too easy for someone just getting started to break the test project by changing a file that isn't obviously connected. Especially as the developer is told to change/replace these assets in the main project but not in the test one--where they don't really matter.

I can see why you might want to avoid duplicating the resources but there is a risk of causing confusion/problems down the line and it's only 19KB that are being saved so it's not a big saving in disk size.
One of the original principles guiding the initial creation of the Test project templates was to mimic what is created by the test project templates that ship with VS. Making this change would break that. Not a big deal but then I'm not sure this brings a benefit either.


If going to accept this change, the equivalent should also be done for the XUnit equivalent project.

sibille:
Hi @mklemarczyk, thanks for your contribution!

I'd like to better understand the motivation for the change you propose. 
Could you open an issue so we can discuss the change you propose? This will also help us to track the change and identitfy other templates affected (as @mrlacey stated). 

We've set up a small contribution guideline doc with more details: https://github.com/microsoft/WindowsTemplateStudio/blob/dev/CONTRIBUTING.md

sibille:
Closing this as it needs further discussion first

mklemarczyk:
It is simply required by builder. The assets need to be exact same files. Otherwise the test project will not work at all.

Issue #3687, opened 2020-05-27
Translation error in code/src/UI/Resources/StringRes.zh-CN.resx and StringRes.zh-TW.resx


kxxt:
**Describe the bug**
Command line should not be translated
```
File StringRes.zh-CN.resx 
  <data name="ErrorMissingNugetPackagesInstallTemplate" xml:space="preserve">
    <value>Install-Package -ID {0} -版本 {1}</value>
    <comment>{0}: package id, {1}: version</comment>
  </data>
```
```
File   StringRes.zh-TW.resx
 <data name="ErrorMissingNugetPackagesInstallTemplate" xml:space="preserve">
    <value>安裝封裝 ID {0} - 版本 {1}</value>
    <comment>{0}: package id, {1}: version</comment>
  </data>
```

**Additional context**
The bad translation will show up in "ERROR_NugetPackages_Missing.txt".

sibille:
Merged now, thanks again for your contribution!

PR #3688, opened 2020-05-27
Do not translate command line.

kxxt:
#3687


sibille:
Hi @kxxt, great catch, thanks for the contribution! Not sure why the build is failing, does not seem to be related to your change!

kxxt:
Hi @sibille. I'm also wondering why changing some resource strings would cause the build failure.

sibille:
Must be sth else, checking the build machine...

sibille:
Fixed! We recently updated the template engine dlls, and had to refresh the cache of the test templates

Issue #3698, opened 2020-06-03
Login feature makes app crash


LoriBru:
I am using version 3.6.20139.1 and Visual Studio 16.6.1.
Creating an application with the Login feature (either forced or optional) makes the application crash at startup.

![image](https://user-images.githubusercontent.com/22811725/83603930-90f5b580-a575-11ea-8dde-61eeb734ad40.png)

sibille:
@LoriBru, thanks for reporting!
Could you give more details on the framework/proyecttype you used to create the application and what's the exact error on app failure?

Did you change the IdentityClientId on the App.config file following the instructions from the IdentityService class ? 

        // WTS TODO: Please create a ClientID following these steps and update the app.config IdentityClientId.
        // https://docs.microsoft.com/azure/active-directory/develop/quickstart-register-app

LoriBru:
I think you're right, that's likely to be the issue.
I think, though, that it would be better to try/catch the "PublicClientApplicationBuilder.Create" call and throw an exception directly from there. It would be a much more precise and informative hint to what is missing.

sibille:
@LoriBru, just saw that the "WTS TODO" comment should be "TODO WTS", that's why it's not showing up in the TODO list, will fix this. 

Regarding wrapping the PublicClientApplicationBuilder.Create, I think the exception thrown from the PublicClientApplicationBuilder is already including all necesary information: 
![image](https://user-images.githubusercontent.com/24698018/83859652-84609100-a71e-11ea-8dda-a2fe75c90923.png)

sibille:
Also opened #3709 to add a global exception handler to UWP templates

mvegaca:
Verified in dev nightly
Templates version: 0.22.20160.2
Wizard version: 0.22.20160.1

PR #3711, opened 2020-06-05
Correct TODO's

sibille:
- Quick summary of changes
Correct TODO's
#3698 Login failure

### DIFF
--- a/code/test/Templates.Test/CodeStyleEnforcementTests.cs
+++ b/code/test/Templates.Test/CodeStyleEnforcementTests.cs
@@ -105,15 +105,15 @@ namespace Microsoft.Templates.Test
         [Fact]
         public void EnsureCodeDoesNotUseInvalidTodoCommentIdentifier()
         {
-            void EnsureUwpTemplatesNotUsed(string fileExtension)
+            void EnsureWTSTODONotUsed(string fileExtension)
             {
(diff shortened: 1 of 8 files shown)


code/test/Templates.Test/CodeStyleEnforcementTests.cs
```diff
@@ -102,6 +102,20 @@ void EnsureUwpTemplatesNotUsed(string fileExtension)
             EnsureUwpTemplatesNotUsed("*.vb");
         }
 
+        [Fact]
+        public void EnsureCodeDoesNotUseInvalidTodoCommentIdentifier()
+        {
+            void EnsureUwpTemplatesNotUsed(string fileExtension)
```
mrlacey, line 108: This method name is not indicative of what it does

sibille: How about EnsureCodeDoesNotUseSwappedTodoComment ?
I want to test that I do not accidentially swap TODO and WTS.

mrlacey: Grrrr, GitHub interface. My comment was against line 108. "EnsureUwpTemplatesNotUsed" is what isn't appropriate

sibille: ah, ok, now I see it, changed!

Issue #3742, opened 2020-06-22
Add two instances of each template that is multiInstance to stylecop tests


sibille:
Follow up work from #3704. Stylecop tests should add two instances of templates that allow multi-instance, to ensure generated code is still valid and correctly formatted.

mrlacey:
Is it necessary to test adding both a single instance of a template and multiple (two) copies?
I don't think there's a scenario where having one instance might pass, but two will fail (or vice versa) but wanted someone else to have thought about this too.

sibille:
we have the one by one tests that are run on a weekly basis that cover the case of a single page being added on each projecttype/framework. This occasionally fails if usings/nuget packages are not added correctly from the package.

sibille:
As this has some impact on test duration (aprox 2 min per framework) and we have two sets of stylecop tests (Group1 and Group2), I've added two instances only on G2 (which includes BackgroundTask which is multiinstance) tests

sibille:
Verified in dev

PR #3758, opened 2020-06-25
Add multiple instances of every multiinstance template to StyleCop tests

sibille:
Quick summary of changes
- Add multiple instances of every multiinstance template to StyleCop G2 tests

#3742 Add two instances of each template that is multiInstance to stylecop tests

### DIFF
--- a/code/test/Templates.Test/BaseGenAndBuildFixture.cs
+++ b/code/test/Templates.Test/BaseGenAndBuildFixture.cs
@@ -60,11 +60,16 @@ namespace Microsoft.Templates.Test
             return userSelection;
         }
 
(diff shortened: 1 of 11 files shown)


mrlacey:
In that this was prompted by an issue with the VB templates, why does this only adjust the tests of the C# code?

sibille:
It includes VB Style tests too, anything else that should be included?

mrlacey:
> It includes VB Style tests too, anything else that should be included?

sorry. missed that

sibille:
No problem, I almost missed them ;)
I think it would be a good idea to include them in the BuildTemplateTests for UWP, to have them next to the stylecop tests.

Issue #3762, opened 2020-06-25
Fix warning about missing Mono.Posix refrence


mrlacey:
The source reports a (10!) warning(s) about being unable to find a reference to Mono.Posix.
Warnings should be fixed on the broken windows principle, and because they risk obscuring other warnings.

This warning is due to a dependency included in CoreTS so that it can run on platforms beyond Windows.

mrlacey:
Error description is:

> Could not resolve this reference. Could not locate the assembly "Mono.Posix". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.

mrlacey:
This affects some projects in CoreTemplateStudio too.

mrlacey:
> This affects some projects in CoreTemplateStudio too.

If I add appropriate missing references to get this building on Windows without warnings, including with a conditional OS check, how do I verify that this has no negative impacts on CoreTS directly, or on WebTS?

sibille:
We need to verify the changes on Mac and Linux, to ensure that local development of WebTS continues to work.

I think there is an alternative to adding Mono.Posix to the WinTS projects. 
Mono.Posix is only used in local development (+ only on WebTS + never in Windows), so according to  https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files we should be able to not expose this package to projects that consume CoreTemplateStudio setting "Private assets" to "all" on Mono.Posix in CoreTS. 
Adding an OS condition for this package might also solve the problem. 

Checking possible solutions on Linux right now... 

Also, can we move this to milestone 3.8? I agree we should fix this, but it's nothing that impacts the final vsix and we're close to code complete for 3.7.

mrlacey:
moved to 3.8  - just want this to not be showing warnings for me. ;)

mrlacey:
@sibille Can you transfer this to be a CoreTS issue and address there? (I don't have permission to transfer, nor the set-up to test impact on WebTS.)

PR #3763, opened 2020-06-25
Mono.Posix warnings

mrlacey:
For #3762 

This fixes the warnings for projects in WinTS but not the projects from CoreTS.

Need to know how to deal with the projects from CoreTS https://github.com/microsoft/WindowsTemplateStudio/issues/3762#issuecomment-649757387


mrlacey:
This isn't the best fix for this issue.
Closing in favor of changes in CoreTS

Issue #3784, opened 2020-07-20
Clicking on NavigationView's Settings item throws NullReferenceException


YuliKl:
**Describe the bug**
When starting a new app, it's easy to get into a state where the app throws a NullReferenceException. The way an app developer fixes this error is not obvious to someone not familiar with NavigationView.

**To Reproduce**
1. Create a new app using WTS with Navigation Pane project type.
2. Modify the NavigationView on ShellPage.xaml to show Settings: `IsSettingsVisible="True"`
3. Run the app and click on the Settings item
4. App throws error: 
`System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Windows
  StackTrace:
   at Windows.UI.Xaml.Controls.Frame.Navigate(Type sourcePageType, Object parameter, NavigationTransitionInfo infoOverride)
   at WTSApp1.Services.NavigationService.Navigate(Type pageType, Object parameter, NavigationTransitionInfo infoOverride) in C:\Users\yulikl\source\projects\WTSApp1\WTSApp1\Services\NavigationService.cs:line 62
   at WTSApp1.Views.ShellPage.OnItemInvoked(NavigationView sender, NavigationViewItemInvokedEventArgs args) in C:\Users\yulikl\source\projects\WTSApp1\WTSApp1\Views\ShellPage.xaml.cs:line 111`

**Expected behavior**
Ideally, app navigates to a blank settings page. As an alternative, app does nothing.

**Additional context**
This seems in the broad category of issues covered by #1122

**System**

- **VS Version:** 16.4.11
- **WinTS Wizard Version:** 
- **WinTS Template Version:** 
- **Windows Build:** Insider

sibille:
Thanks for reporting @YuliKl . Underlying issues is that the SettingPage is not available, so we cannot navigate there. 

We see two options: 
1. NavigationService checks for invalid PageType and does not navigate.
2. NavigationService checks for invalid PageType and throws a more meaningful exception.
```
if (pageType == null || !pageType.IsSubclassOf(typeof(Page)))
{
   throw new ArgumentException("Invalid pageType, please provide a valid pageType.", nameof(pageType));
}
```
From our team we prefer the second option, as this helps the Dev to detect potential issues in the app, but also see that option 1 makes the app more robust. Would love to hear your opinion.

YuliKl:
Both options seem reasonable. I agree that option two is more informative and therefore preferable.

[...]

PR #3791, opened 2020-07-24
Add PageType validation in UWP MVVM Basic and CodeBehind Navigation S…

mvegaca:
Clicking on NavigationView's Settings item throws NullReferenceException #3784


mrlacey:
Will a separate PR be raised for other frameworks? If verified that not needed for other frameworks this should be documented.

PR #3794, opened 2020-07-25
Replaced MSTest asset files with link to main project.

mklemarczyk:
Quick summary of changes
MS Test project should refer to main project assets files.
The real files was replaced by project link in this PR.

The change should be tested if the links are resolved properly.

For people that do not understand the problem (as mrlacey and sibille). Sorry for no answer on last pull request, but the GitHub is not my life. I just want to simplify the beginner developers life in figuring out why theirs test project stop working when they inserted a new asset files. If new files will be added in main project and do not correspond in test project, the duplicated assets issue will not trigger. If the asset file will be removed in IDE, IDE also removes it from related project. There is no problem with this solution.

If you try to change assets in main project and not in the test project. The MSBuild will than trigger an exception that the asset files are defined twice. You will not be able to continue with running tests anymore. It simply has to be the same files (byte to byte exact). In all other cases the test project will stop working. I was thinking about system symlinks, but it is not easy to achieve and maintain. If you see other options to follow, please go ahead.

I would like to address this issue also with Microsoft team, that is responsible for it. In my opinion test project should not carry about icon/image assets as it is only to verify if the application works.


mrlacey:
Oh, I see. 
Yes, this makes sense.
What was missing in #3599 was a description of the issue the change addresses. This is why the need to first create an issue that explains what's wrong or needs changing is called out in the [contribution guide](https://github.com/mrlacey/WindowsTemplateStudio/blob/dev/CONTRIBUTING.md).

---

Because of restrictions on the build server, there are no automated tests for the code affected by these changes. I have manually verified this all works ok.

- VB changes are also needed. 

- I'm not sure if the underlying issue is due to a [WinUI](https://github.com/microsoft/microsoft-ui-xaml/) restriction or something in [testfx](https://github.com/microsoft/testfx).

sibille:
Thanks for your contribution @mklemarczyk!

I followed your steps and understand the motivation of your change now, thanks for the explanation. 
Same happens using xUnit for for App Tests. I don't think this is related to WinUI as the same error occcurs in a Blank App.

I'll open an issue to track this change for all affected frameworks and languages. 

@mklemarczyk, are you interested in adding the same fix for xunit and vb or prefer the team to jump in here.

templates/Uwp/Testing/UnitTests.App.MSTest/Param_ProjectName.Tests.MSTest/Param_ProjectName.Tests.MSTest.csproj
```diff
@@ -125,13 +125,27 @@
   </ItemGroup>
   <ItemGroup>
     <Content Include="Properties\UnitTestApp.rd.xml" />
-    <Content Include="Assets\LockScreenLogo.scale-200.png" />
-    <Content Include="Assets\SplashScreen.scale-200.png" />
-    <Content Include="Assets\Square150x150Logo.scale-200.png" />
-    <Content Include="Assets\Square44x44Logo.scale-200.png" />
-    <Content Include="Assets\Square44x44Logo.targetsize-24_altform-unplated.png" />
-    <Content Include="Assets\StoreLogo.png" />
-    <Content Include="Assets\Wide310x150Logo.scale-200.png" />
```
MSLukeWest, line 134: Shouldn't these seven items be removed since you're replacing them each with a content item that uses a link?

sibille: @mslukewest, can't see whats wrong here, could you give more details?

MSLukeWest: Ignore my comment, I read the diff wrong and thought all the items had duplicates.  I see now that they don't.

sibille:
Related issue:
#3795 
This PR solves the issue for MSTest C# templates, will do a separate PR for the other affected test frameworks and templates

Issue #3607, opened 2020-04-01
WPF Page: ContentGrid


mvegaca:
Add a **ContentGrid** page template for **WPF**.

Action Plan:

 - [x] Create mockups
 - [x] Product team validation
 - [x] Create templates
 - [x] Add docs

You can find the mockups at https://github.com/mvegaca/ContentGridMockup.
This page contains a basic content grid based on a list view. Selecting an item shows a detail page.
Items rearrange when resizing the window but there's no adaptative behaviour or connected animations like in the UWP Content Grid Page.

![image](https://user-images.githubusercontent.com/24698018/85841574-98883300-b79e-11ea-84e4-337786a8b319.png)
![image](https://user-images.githubusercontent.com/24698018/85841616-a6d64f00-b79e-11ea-9728-3274ea87e8ca.png)

sibille:
Verified in dev-nightly: 
Templates version: 0.22.20252.1
Wizard version: 0.22.20251.1

PR #3802, opened 2020-08-07
WPF Content Grid Templates

mvegaca:
WPF Content Grid Templates #3607


sibille (changes requested):
Texts from UWP Content Grid are not valid for WPF as this is not using the adaptative GridView. 
We should use something similar to the following texts: 
- Short description: This page contains a basic content grid based on a ListView. 
- Long description: This page allows you to add custom content in a to a content grid based on a ListView control. Clicking on an item shows a detail page.

sibille (approved):

Issue #3613, opened 2020-04-01
WPF Testing Templates


mvegaca:
Add Testing templates for WPF, such as:
- Unit tests with XUnit/Nunit/MSTest
- UI Tests using WinAppDriver

Action Plan:

 - [x] Create mockups
 - [ ] Product team validation
 - [x]  Create templates
 - [x] Add docs

First code PoC could be found on [this repository](https://github.com/mvegaca/MyDotNetCoreWpfApp/tree/dev/NewFeatures).

sibille:
Mockups are finished and published in branch https://github.com/microsoft/WindowsTemplateStudio/tree/Mockup-WPF-Testing.

The idea for the testing templates in WPF is the same as for UWP: 
- Provide a basic test class with an empty test 
- Provide sample tests for the pages/features and services that were selected on project creation

If you add a new testing project via right click, you will only get the basic test class, as we cannot know what else is already in the project (you might have added a MasterDetail page, but already removed the SampleDataService).

For WinAppDriver the test setup is different for projects with or without MSIX package  The code in the mockup is valid for both scenarios and has developer TODO instructions on how to adjust the code according to the scenario.


We would love to get your feedback on the mockups.

mrlacey:
Do these tests work ok with projects including a XamlIsland?

mrlacey:
Including the temporary `.pfx` file in the repository will make it easier to review the mockups as this will make it possible to compile the project after downloading the branch

mrlacey:
The following comment isn't clear.

`// TODO WTS this is not necessary if your project doesn't include MSIX`

Is this explaining the following code block? If so it doesn't need to be a `TODO`.
If the intended action here is to remove the following code block make this clearer.

mrlacey:
The test `TestDeleteFile` in `FileServiceTests` relies on one of the other tests in that class having been run first.
This is because this test assumes the existence of a directory that will not exist before the other tests have run. The other tests check for the existence of the directory and create it if it doesn't exist.

This is particularly important as the Test Explorer can list tests by name and the current names will result in the Delete test running first.


![image](https://user-images.githubusercontent.com/189547/94853473-5caadc80-0423-11eb-90ce-bdc05e66ac86.png)

vmadurga:
Hi Matt, thanks for your feedback.

We updated the current mockup branch.

We include all of your suggestions and include a XAMLIsland

[...]

PR #3948, opened 2020-11-11
Fix project references to use ProjectName instead of RootNamespace

sibille:
**Quick summary of changes**
Fix project references to use ProjectName instead of RootNamespace
Detected by full test pipeline https://winappstudio.visualstudio.com/WTS/_build/results?buildId=35473&view=results

#3613 WPF Testing Templates

| No | Yes|No |


mrlacey:
Was this found as part of a failed automated test?
Should tests be added for this? Could the use of RootNamespace elsewhere (or in the future) also cause errors that should be caught with tests?

sibille:
This was detected by the ProjectNameValidation that run on the full test pipeline. 
Other errors caused by using RootNamespace instead of project name should be detected by those tests too.

mvegaca (approved):

Issue #3817, opened 2020-09-02
Unify template names format


mvegaca:
There are some pages with Name and ShortName with white space like "ContentGrid" or "Two Pane View" and others without white spaces like "DataGrid" or "TreeView".

I suggest using the same naming format for all pages in UWP and WPF.
 - Blank
 - Settings
 - Web View
 - Media Player
 - Master / Detail
 - Two Pane View
 - Tree View
 - Content Grid
 - Data Grid
 - Telerik Data Grid
 - Chart
 - Tabbed Pivot
 - Tab View
 - Map
 - Camera
 - Image Gallery
 - Ink Draw
 - Ink Smart Canvas
 - Ink Draw Picture

The naming format are already well implemented on features and services. This modification only applies to json files and not be necessary to get new translations. We will need to add the telemname attribute in template json to keep the telemetry data.

sibille:
We should use the same name as the underlying control name if there is one, f.ex WebView and TwoPaneView.

vmadurga:
Proposal page names using the underlying control name:

| Page name | Underlying control |
| ------------- | ------------- |
|Blank| - |
|Settings| - |
|WebView | WebView |
|MediaPlayer | MediaPlayerComponent |
|MasterDetail | MasterDetailsView |
|TwoPaneView | TwoPaneView |
|TreeView | TreeView |
|Content Grid | AdaptiveGridView |
|DataGrid | DataGrid |
|Telerik DataGrid | RadDataGrid (Telerik) |
|Chart | RadCartesianChart (Telerik) |
|Tabbed Pivot | Pivot |
|TabView | TabView |
|Map | MapControl |
|Camera | CaptureElement |
|Image Gallery | GridView |
|Ink Draw | Canvas/InkCanvas |
|Ink Smart Canvas | Canvas/InkCanvas |
|Ink Draw Picture | Canvas/InkCanvas |

mrlacey:
Showing existing names and the proposed new ones would be more useful than documenting underlying controls

mrlacey:
Is there a definitive list of all the names and what they should be?

Need to check the templates for the MVVM Toolkit (#3809 & #3753) use the correct names

sibille:
Please take the names from the "Proposal page name" column in https://github.com/microsoft/WindowsTemplateStudio/issues/3817#issuecomment-706159223

sibille:
Verified in dev-nightly version: 
Templates version: 0.22.20335.2
[...]

PR #3953, opened 2020-11-12
Unify template names in all platforms

mvegaca:
# PR checklist

**Quick summary of changes**
Unify template names in all platforms.

Unify template names format #3817
 Open

| Yes | Yes | Yes |


sibille:
Media Player should be changed to MediaPlayer and TabView should not be changed but keep called TabView.

sibille (approved):

Issue #3910, opened 2020-10-20
Cleanup temp generation folder on SetAgentDriveAndCleanTestFolder


sibille:
Templates tests fail from time to time on cleaning up the temporary generation folder.
We should cleanup that folder from the SetAgentDriveAndCleanTestFolder script to avoid tests failing here.

sibille:
Verified in build pipelines

PR #3956, opened 2020-11-17
Update SetAgentDriveAndCleanTestFolder.bat

sibille:
**Quick summary of changes**
Clean temp generation folder before running tests on the build pipeline to avoid errors from concurrent test execution.

#3910


prupipho (approved):
Not sure if I can approve this. But looks perfect to me 👍🏻 Do we have to do this on WebTS as well to remove template_test folder?

sibille:
You can :) We don't have this for WebTS I think, I'll check and open an issue if it's not there yet.

PR #4103, opened 2021-02-26
WinUI3 AppModels Tests

mvegaca:
# PR checklist

**Quick summary of changes**
Adapt tests to run WinUI 3 app tests in both different app models (UWP and Desktop)

Add support for AppModel filter in templates and composition queries
https://github.com/microsoft/CoreTemplateStudio/issues/287

| No | No | Yes |


sibille:
Split WinUI 3 Tests to both app models included in PR #4120

Issue #4100, opened 2021-02-25
Move our templates higher up in the New Project Dialog template list


sibille:
See if we can get higher in the New Project Dialog template list by setting SortOrder
https://docs.microsoft.com/en-us/visualstudio/extensibility/sortorder-element-visual-studio-templates?view=vs-2019

Applies to the following platforms:
| UWP              | WPF              | WinUI           |
| :--------------- | :--------------- | :---------------|
| No | No | Yes |

mrlacey:
Why is this only being done for WinUI and not also for UWP & WPF templates?

Why was this originally configured with a sort order that puts it very low?

What's driving the change? Is there a specific problem or piece of feedback that has led to this change?

sibille:
Cannot see any difference in the dev-nightly version by setting the sort order.

sibille:
Moving to milestone 4.1 while following up internally on more guidance on this

mikebattista:
@mrlacey we should confirm the behavior here with the VS 2022 extensions.

mrlacey:
@mikebattista assuming that no response was received from previously "following up internally", what is your goal or expectation here?
Is it just to have the entries appear as high as possible in the New Project list? Or do you have ideas for specific placement?

mikebattista:
Do you know for sure no response was received? We should confirm the behavior of increasing the SortOrder to make the Template Studio templates more visible. We certainly don't want the SortOrder at the default value if everyone else is bumping up the priority of their templates.

mrlacey:
> Do you know for sure no response was received? We should confirm the behavior of increasing the SortOrder to make the Template Studio templates more visible. We certainly don't want the SortOrder at the default value if everyone else is bumping up the priority of their templates.

I assume no response based on the last [publicly] documented statement being https://github.com/microsoft/TemplateStudio/issues/4100#issuecomment-838121636

mikebattista:
I doubt there was closure and there's no record of it, so we should confirm the behavior.

mrlacey:
They're configured to be as high as they can go.
VS2022 changed some of the behavior around the ordering and so templates from extensions are now given lower priority. :(

mikebattista:
Ok. I noticed they were all set to 10 already as well. I guess we can close this out.

PR #4111, opened 2021-03-02
Move our templates higher in the New Project Dialog list

sibille:
**Quick summary of changes**
Setting sortOrder to move our templates higher in the New Project Dialog list

#4100 Move our templates higher up in the New Project Dialog template list

| No | No |Yes|


mrlacey:
Why the choice to put them as high as possible on the list?

It there any existing etiquette or precedence for defining where in the list to put entries?

sibille:
Docs say that this affects templates in the same category. 
For the moment this PR is to see if this causes any effect on the templates in the dev-nightly.

mvegaca (approved):

PR #4115, opened 2021-03-04
Update with data from February

sibille:


harvinders:
Seems like European decimal symbol `","` is replacing `"."`. Is it deliberate?

sibille:
Great catch, thanks @harvinders! Will change this back.

prupipho (approved):

Issue #4029, opened 2021-01-12
Support Non-navigating NavigationViewItems


myokeeh:
@mvegaca, I think I'm recalling why I did this. It was to account for cases where I didn't want navigation to occur for certain  `NavigationViewItem`s. One case is when you have nested `NavigationViewItem`s where for the top/parent, `helpers:NavHelper.NavigateTo` is NOT set because it's more of a container for children `NavigationViewItem`s

It appears even onthe latest version (wizardVersion: v3.9.20351.2, templatesVersion: v3.9.20351.2), clicking on such an `NavigationViewItem` will throw exception below.

>throw new ArgumentException(string.Format("Invalid pageKey '{0}', please provide a valid pageKey. Maybe you forgot to call NavigationService.Configure?", pageKey), nameof(pageKey));

How is WTS going to account for such a case?

_Originally posted by @myokeeh in https://github.com/microsoft/WindowsTemplateStudio/issues/3806#issuecomment-754970442_

mrlacey:
It sounds like you should be using a `NavigationViewItemHeader` instead of a `NavigationViewItem`.

![image](https://user-images.githubusercontent.com/189547/104362364-a79cbb80-550b-11eb-9d7d-990dbf8f993f.png)

![image](https://user-images.githubusercontent.com/189547/104362327-9b186300-550b-11eb-8d81-29d0da1b8c9b.png)

If for some reason you can't do this, you could have the check for a page parameter not throw an exception and instead do nothing.

![image](https://user-images.githubusercontent.com/189547/104362656-fe09fa00-550b-11eb-9239-ff059ce50cce.png)

myokeeh:
`NavigationViewItemHeader` doesn't look like it's collapsible the way you'd be able to if you nest `NavigationViewItem`s. I've removed the Exception throw as you've pointed out.

mrlacey:
> `NavigationViewItemHeader` doesn't look like it's collapsible the way you'd be able to if you nest `NavigationViewItem`s.

It sounds like you want [Hierarchical Navigation](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#hierarchical-navigation) rather than repurposing individual items.

sibille:
@myokeeh, does Hierarchical Navigation with NavigationView.MenuItems solve your issue?

myokeeh:
[...]

PR #4202, opened 2021-04-09
Support Non-navigating menu items in split view projects

dgomezc:
# PR checklist

**Quick summary of changes**
 - Support Non-navigating menu items in UWP split view projects (C# and VB).
 - Support Non-navigating menu items in WinUI split view projects (C#).

#4029 - Support Non-navigating NavigationViewItems

| Yes | No|Yes |


mrlacey:
As this change adds code to handle scenarios not covered in the code generated from WinTS templates, what testing can be added to verify that the code changes here work, and continue to work with any future alterations to the templates?

sibille:
Not sure how we can test this in an easy way, as it requires modifying the generated project and navigate between items.

Also, this PR is only adding null checks to support this scenario, the existing behavior should not change.

sibille (approved):

Issue #3928, opened 2020-10-27
Need official designs for deprecation flags


mrlacey:
#3809 adds the ability to add "flags" to the frameworks that are being deprecated with the introduction of templates using the MVVM Toolkit.

The implemented designs are purely a placeholder.

![image](https://user-images.githubusercontent.com/189547/97327345-c8d6ff80-186c-11eb-9169-db3db3b6f247.png)

This new UI should be reviewed and appropriately adjusted accordingly.

_This is a follow-on from #3753_

mrlacey:
The above image is placeholder only.
The text is deliberately at the top to make it hard to miss.
The text overlays the image to avoid people making a selection based on recognizing the image and not looking at anything else. Obscuring the image was deliberate.
The inverted text (background and foreground colors) is to make the deprecation prominent.
Even if someone never uses the deprecated options, the deprecation notice is to make all users aware that a change is coming. This should minimize the surprise when these options are removed.

sibille:
I looked around a bit on other apps in the ecosystem to see how they mark things deprecated (or in preview). Leaving a few screenshots for reference here.

Windows Community Toolkit:
![image](https://user-images.githubusercontent.com/24698018/114197438-a33d3e80-9952-11eb-824c-817564443f72.png)

XAML Control Gallery:
![image](https://user-images.githubusercontent.com/24698018/114197778-f7482300-9952-11eb-9409-d10354302d1d.png)


Visual Studio: 
![image](https://user-images.githubusercontent.com/24698018/114197535-be0fb300-9952-11eb-9acf-382f6db01b22.png)

sibille:
Design Team recommended us to:
- Not cover the framework images as some of them are brand images
- Use the warning icon for deprecated frameworks as the Nuget Package Manager does
- Show the warning icon and deprecated text below the framework name 
- Make sure "deprecated" text is pronounced by screen-readers

Card content would be similar to: 

**MVVM Light**
⚠️ Deprecated
MVVM Light is a ....

sibille:
Verified on dev-nightly: 
Templates version: 0.22.21110.1
Wizard version: 0.22.21110.1

PR #4209, opened 2021-04-14
Add deprecated flag

dgomezc:
# PR checklist

**Quick summary of changes**
  - Add deprecated message in framework cards.
  - Add deprecated message in framework detail pages.
 
#3928 - Need official designs for deprecation flags

| Yes | Yes | No |

Sscreenshots:
Framework selection page
Frawemork detail page


mrlacey:
When I added the original support for a deprecated flag, it was done with the flexibility that it could be used for any text or future "flag" that may be needed".
This change appears to have hard-coded the warning icon but this may not be appropriate for all future "flags".
Was making the icon configurable considered?

I'm ok if the answer is "we decided against this but could add it in the future if needed."
What I'm interested in is the decision-making process and the options that were considered. There is no documentation of the process or the reasoning behind any decisions. I have ideas for future functionality that will require the use of different "flags" and now icons. If work has been done that will make this easier or harder or investigations have discovered things (such as edge-cases) that need to be considered as part of an implementation it will save time in the future (regardless of who implements them or why) if these are documented now.

dgomezc:
We only requested feedback from the design team for the deprecated flag. We are not sure if the same design is valid for other flags. 

We used "Flag" property because is was already created but we think this should use a dedicated "IsDeprecated" tag. If we add other flags we should ask the design team for the correct way to show them and adapt the code to include those flags.

code/src/UI/ViewModels/Common/DataItems/FrameworkMetaDataViewModel.cs
```diff
@@ -33,12 +33,12 @@ public FrameworkMetaDataViewModel(MetadataInfo metadataInfo)
             Order = metadataInfo.Order;
             MetadataType = metadataInfo.MetadataType;
             Licenses = metadataInfo.LicenseTerms?.Select(l => new LicenseViewModel(l));
-            Flag = GetLocalizedFlag(metadataInfo.Tags.FirstOrDefault(t => t.Key.Equals("wts.flag")).Value?.ToString());
+            Deprecated = bool.TryParse(metadataInfo.Tags.FirstOrDefault(t => t.Key.Equals("deprecated")).Value?.ToString(), out bool isDeprecated);
         }
 
-        private string GetLocalizedFlag(string flagValue)
+        private string GetLocalizedTag(string flagValue)
```
mrlacey, line 39: Where is this used?
What are the "tag"s that this relates to?

mrlacey:
What's the reason for this latest change?
The flag capability was originally created to be flexible because I knew this would be reused for other things in the future.
I know (because you've said above) that designs are only approved for the deprecated flag but I don't see why the existing flag capability could not have been used for this.

With this change is looks like:
- You've done extra work for no immediate benefit.
- You've removed the capability for future flexibility, meaning that additional work will be required in the future when we need a flag for other things.


I'm assuming there's something I'm missing in the above decision-making. Please fill me in.

sibille:
I actually liked the flexibility of the flags property, but the design proposal we got from the design team is very specific to flagging a deprecation, that's why we decided to create a specific tag for this.
We were thinking about leaving the flags property, but as we currently do not have any flags we decided to remove it (for now). What other flags are you thinking about? Preview, New?

sibille (approved):

Issue #4301, opened 2021-05-20
Review Code suggestions on Wizard


vmadurga:
Review code suggestions on Wizard, such as:

- Make fields readonly
- Use pattern matching
- Simplify object initializations
- Redundant and unnecesary assignments
- Simplify interpolations
- Do not use Count when Any can be used
- Use nameof to express symbol names
- Deconstruct variable declarations
- Use Count/Length property instead of Count()
- Simplify conditional expressions
- Remove unnecessary suppressions
- Remove unnecessary assignments
- Remove unused methods
- Remove unused function parameters
- Remove not access fields
- Offer async method options
- Remove unused directives

sibille:
Verified in dev-nightly:
Templates version: 0.22.21167.1
Wizard version: 0.22.21167.1

PR #4329, opened 2021-06-01
Fix release build

sibille:
# PR checklist

**Quick summary of changes**
Add usings for release configuration

#4301 Review Code suggestions on Wizard

### DIFF
--- a/code/src/UI/VisualStudio/RightClickActions.cs
+++ b/code/src/UI/VisualStudio/RightClickActions.cs
@@ -19,6 +19,11 @@ using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.TemplateWizard;
 using Microsoft.VisualStudio.Threading;
 
+#if !DEBUG
+using System.Reflection;
+using Microsoft.Templates.Utilities.Services;
+#endif
+
 namespace Microsoft.Templates.UI.VisualStudio
 {
     public class RightClickActions : IContextProvider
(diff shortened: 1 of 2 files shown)


mrlacey:
Why are these using directives excluded from the DEBUG build?

If these really only must be included in the Release build then a comment in the code to explain why will be helpful for anyone coming across this in the future.

Also, how was this discovered? 
That (according to the OP) this change was made in response to a code suggestions is surprising. Missing using directives normally create more than a suggestion. 
If this was to simplify code elsewhere in the file why no changes anywhere else?
If this was because the code didn't compile in release mode then more than this change is needed. Is the code not compiled in the release configuration as part of the CI build verification?

mrlacey:
These new comments explain what the code is doing (by being behind a compiler directive) it doesn't explain why these directives aren't needed in the debug build so I looked at the whole file and saw other compiler directives there.

I'm going to stand by my second assumption then that the code doesn't build in Release mode and so the bigger concerns of:
- how/why this wasn't tested by the devs who made and approved the change; and
- that the CI build isn't checking the release build of the software 

means it's worryingly easy for code that doesn't compile to be merged and for this only to be discovered when preparing for a release.

sibille:
The directives are only necessary for the release configuration, they were accidentally removed cleaning up unused usings in the issue mentioned. 
I added them back only for release to make it clearer that they are only used in this scenario. 

This was detected by the pipeline that creates the dev-nightly version, I will review the pr and ci build pipeline to prevent this from happening again.

Also added the comment you suggested.

dgomezc (approved):

PR #4392, opened 2022-01-26
Converted ADO Pipelines to YAML

mikebattista:
Checked in existing CI pipelines as YAML so we have a baseline to work from going forward.

### DIFF
--- a/_build/pipelines/ci.yml
+++ b/_build/pipelines/ci.yml
@@ -1,0 +1,100 @@
+name: $(date:yyyyMMdd)$(rev:.r)
+
+trigger:
+  branches:
+    include:
+    - dev
+    - release
+  batch: True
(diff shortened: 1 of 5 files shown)


mrlacey (approved):

mrlacey:
In that some of the PR checks have failed, it suggests that the triggers for this aren't right.
It shouldn't be running everything for a PR

mikebattista:
Most of these are older runs I manually triggered during testing. The only one that should be required is the "PR" test.

Once everything is merged I'll also need to change the required status check from (old) wints.pr to (new) Template Studio PR.

mikebattista:
Both the old and new PR pipelines passed. The status check reporting looks confused. I'm going to merge.

Issue #4353, opened 2021-06-17
Visual Studio 2022 Support


scottkuhl:
### Installed product versions
- Visual Studio: 2022 All Versions

### Description
Make the extension available in Visual Studio 2022.

### Steps to recreate
1. Download the extension installer.
2. Run the installer.

### Current behavior
Message that it is already installed to all applicable products appears even though it is not installed in Visual Studio 2022.
It is also not listed in Online Extensions in Manage Extensions Visual Studio 2022.

### Expected behavior
Installs and works in Visual Studio 2022.

_Note: Sorry for the all the extra details.  I probably could have stopped with the title on this one._

scottkuhl:
[Writing Visual Studio Extensions with Mads - Splitting extensions for Visual Studio 2022](https://www.youtube.com/watch?v=-PKIPTW6km0)

sibille:
Thanks for opening this issue. We've looked into support for VS 2022 but need a few architecture changes before being able to support VS2019 and VS 2022 side by side.

itsbeer:
Any updates on this? 

Its my understanding that there are libraries that have x64 architecture issues that are being waited on before an updated extension can be released. Is that still the situation? If so, what exactly are the libraries at issue? 

With that knowledge, maybe the community at large can help speed things along. This is hands down one of my favorite and most useful extensions. It would be a shame for the VS2022 user base not to benefit from it.

mikebattista:
I'm digging into this at the moment. Whatever remaining blocking issues I find, I will list them here.

mrlacey:
> I'm digging into this at the moment. Whatever remaining blocking issues I find, I will list them here.

Here's a new issue to add to the list: https://developercommunity.visualstudio.com/t/creating-project-dialog-is-always-on-top-in-vs2022/1608136

mikebattista:
Thanks. I did get things running in the latest 2022 preview and noticed this bug as well. I'm working in [my fork](https://github.com/mikebattista/WindowsTemplateStudio) if you want to take a look. It needs more validation.

[...]

PR #4408, opened 2022-03-01
Refactor and restructure for VS2022 support

mrlacey:
# PR checklist

**Quick summary of changes**

This is where the changes to support VS2022 will appear. Please raise comments, questions, etc. as appropriate.

More details will follow.

For #4353 and [this milestone](https://github.com/microsoft/TemplateStudio/milestone/33)

| UWP              | WPF              | WinUI            |
| :--------------- | :--------------- | :----------------|
|  Yes | Yes | Yes |

The primary goals of the changes included here are:
- Support Visual Studio 2022 (& drop support for earlier versions of VS)
- Break the project up so that there are separate installers for **UWP**, **WPF**, **WinUI (C#)**, and **WinUI (C++)** versions of the wizard & templates.
- Simplify and standardize the codebase to make it easier for others to work on/with. (While much work has been done towards this goal it's a long-term goal and there will always be opportunities for improvements.)


mrlacey:
Using a draft PR that I intend to add to bit-by-bit so that changes are visible and I can add comments where appropriate.
This is instead of just submitting all changes in one go.
Hopefully, this will make the changes easier to understand. It also avoids me accidentally including things that shouldn't be added here if I brought in my entire experimental fork.


**Note.** I expect that the automated tests will break for this PR due to extensive internal changes to the structure of the code, the changes in the projects included in the solution, and that there will now be 4 projects producing output rather than just one.

mrlacey:
**Explaining the solution/project structure.** ([Visible here](https://github.com/microsoft/TemplateStudio/issues/4353#issuecomment-1057143446))

The basic principle is that each extension has 3 parts:
1. The project that produces the VSIX extension. (Named `TemplateStudioForXxxx`)
2. The project that hosted the "Project Template" the extension project references. This is what adds the entries in Visual Studio's New Project wizard. (Named `Ts4XxxxTemplate`.) Note that the two UWP entries are used in one extension project while the WinUI ones are in different extensions to allow for separate workload dependencies.
3. The project that has the tests specific to the templates in each extension. (Named `TemplateStudioForXxxxx.Tests`)

Additionally, there are projects for:
- Shared functionality relating to the UI of the wizard (Named `SharedFunctionality.UI`)
- Shared functionality relating to the "Core" logic relating to working with templates. (Named `SharedFunctionality.Core`) This was previously loaded from https://github.com/microsoft/CoreTemplateStudio -- so no more submodule use
- Tests projects for the functionality in each of the above.
- A `SharedResources` project that is recompiled for each of the extensions projects.

There are also some files that are linked (or copied) in multiple projects. (e.g. the colors package, and the contents of the "Styles" folders.) 

I tried to have as few projects and shared/linked files as possible and this is the best I could manage. I encountered obscure issues when working with some files inside Shared Projects that affected the ability to correctly include files within the extension so they could be used at runtime. There were also issues with having a predictable path that could be referenced at runtime in a separate assembly. Basically, if you're wondering why things are as they are it's because this is the simplest thing I could get to work. I know there are lots of things that could be improved. If you want to try changing them, you're welcome to try but I suggest doing it after this is merged. I'm also reluctant to discuss or help with any specific changes or improvements, as I have already spent months working on this and will benefit from a break. That said, there are things I'll definitely try and improve in the future but my attempting to do them now will add an avoidable delay.

mrlacey:
## Configurations!

There's a new solution file (`/code/TemplateStudio.sln`) that contains everything. Consider this a replacement for "Big.sln"

When working with this solution there are two important things to remember:
1. Be sure to choose the correct configuration.
2. If you don't need everything (for instance if you're working with the templates for only one extension) you should use a solution filter.

Let me expand on those two options.

### 1. Choose the correct configuration

To allow for the extensions to be installed side by side, some of the code logic needs to use different internal identifiers and names. These are controlled through conditional compilation. To be sure that the correct values are used for each project, use the appropriate configuration option. (Please say if you don't think the names are clear.)

image

When you use one of these configurations it will only build the projects that relate to that configuration. (This makes building hte whole solution faster as it means you don't need to build the things you aren't interested in.)

> Aside: if you [also] think it's annoying that you can't see the full name of the selected configuration in the dropdown, [please upvote this issue](https://developercommunity.visualstudio.com/t/Active-Solution-Configuration-combo-drop/1297733).

**If you don't chose a specific and appropriate configuration, YOU WILL GET BUILD ERRORS (and possibly warnings as well.)**

image

The default `Debug` and `Release` configurations have been left in the solution for two reasons:
- To catch out anyone who doesn't know about or use an appropriate configuration. I didn't want to default to something with only a 1-in-4 chance of being what's wanted. Better to force a person to choose the configuration they want rather than use something that is likely to be wrong and be unexpected and hard to debug.
- I have in the past run into issues with tooling that breaks when it can't find either the "Debug" or "Release" option. While these should never be used when working with this codebase I don't want anyone to get weird errors that aren't obvious to explain or understand.

If you want to be able to build _and test_ everything, you can use the `DebugALL` configuration as this will ensure that everything is built and runable. Note that this was primarily intended for use when wanting to run all tests. Do not use it for manually testing/debugging code. You may get weird results (or errors) because the SharedResources library may not be compiled correctly (as expected).

There is no "ReleaseALL" configuration as the SharedResources library **_MUST_** be compiled for the extension it is to be used with.

### 2. Using Solution filters

While the configurations (mentioned above) avoid compiling more projects than are necessary they do still get loaded.
If you're only working on one extension (or the templates for one extension) then you will find things faster and easier if you open a solution filter file instead. 
These are a filtered subset of all the projects in the solution and there is one available for each of the produced extensions.

image

Open and use one of these files if you don't need everything.    
**BUT you must still choose the correct configuration!**

---

A version of the above will also, eventually, appear in the "working with the codebase" docs.

mrlacey:
## Localization

There are three places containing localized content:
1. String resources (in the SharedResources project) - these are a combination of all the separate RESX files that existed previously.
2. The `.vstemplate` files for each of the Project Templates.
3. The `Extension.vsixlangpack` files for each of the extension projects.


Combining all the resx content into a single location should make, finding, editing, and localizing these entries easier.

The content of the vstemplate & vsixlangpack files will need updating as they are based on what was used for the previous single installer.

mrlacey:
## Version numbering

All visible version numbers (_should_) come from the extension manifest. (There's only one place to set/change/update this when making a release.)
Templates no longer have a separate version number.
The extension projects have different internal version numbers (specified in AssemblyInfo.cs) that are used as part of the strong name identification from the project template. Having these version numbers be different helps when/if there are problems resolving internal references. When the version number isn't as expected the incorrect value that is being used can help point to where the misconfiguration is.

mrlacey:
## Templates!

They've moved.
They've been split up.
They're included in the projects in a new way.
The code looks for templates in a new way.

### Template location changes

Previously, all templates were in a directory under the root of the project. 
They are now included directly in the extension projects.

The templates used explicitly as part of testing are now in the "TestData" folder of each platform-specific test project.

### Template separation

As there are now separate extensions for the separate target platforms (and development languages) the templates have been divided up based on the extension they will work with (be used by).

When all templates were kept together they were divided into separate directories based on the platform name. The platform named directories have been retained _for now_ (to avoid unnecessary but far-reaching changes when so many other parts are also moving) but should be removed in the future.

### Template inclusion in the extension

Previously, all templates were combined into a single, signed, zip file and included in the extension, and also made available for download separately. 
They're now included directly in each extension.
To ensure they are packaged correctly, they must have a **Build Action** of **Content** and be set to be **included in the VSIX**. No other attributes, actions, or settings should be applied to the template files.
There are tests that ensure that these files are configured correctly in the extension projects.

### Referencing template locations

Previously, the code worked with a `LocalTemplatesSource` or a `RemoteTemplatesSource`. The local version worked with the filesystem directly. The remote version would download, and extract a compressed bundle of published templates.

The code now uses a `VsixTemplatesSource` to work with the templates inside an extension.
There are also multiple different classes that now inherit from LocalTemplatesSource that are used for different types of testing.

Additionally, there used to be many compiler directives to choose between using the LocalTemplatesSource while in Debug and the Remote version in a release build. The code now passes a `TemplateSource` object around so that only what's needed is created and dependency logic doesn't need to be duplicated or spread across multiple files.

Eventually, RemoteTemplatesSource can be removed. (But it hasn't been yet.)

mrlacey:
## VSEmulator

Having to launch an experimental instance of Visual Studio every time you want to test changes to templates can be slow and not a good use of your time. 
VSEmulator is a separate executable that makes it much quicker and easier to launch the Template Studio wizard so you can test changes to templates. 
It can obviously also be used to test changes in the wizard too.

mrlacey:
WPF WACK tests confirmed to all be working
image

Yes, I left this running overnight as each took about 39 minutes to run!!! - (Much longer than expected and I don't know why it's that long.)

mrlacey:
## Test Traits

Tests are classified using traits.

The traits of interest are `"Group"`s

The following groups are defined:

_Tests that should be run a part of a PR as they relate to core functionality and execute quickly_
**Minimum** - Minimum tests that are not specific to any specific generated template.
**MinimumUWP** - Minimum tests that relate to the UWP output
**MinimumWPF** - Minimum tests that relate to the WPF output
**MinimumWinUICS** - Minimum tests that relate to the WinUI (C#) output
**MinimumWinUICPP** - Minimum tests that relate to the WinUI (C++) output

_Release tests are intended to only be run prior to a release. _
**Release** - Release tests that are not specific to any specific generated template.
**ReleaseUWP** - Release tests that relate to the UWP output
**ReleaseWPF** - Release tests that relate to the WPF output
**ReleaseWinUICS** - Release tests that relate to the WinUI (C#) output
**ReleaseWinUICPP** - Release tests that relate to the WinUI (C++) output

**ManualOnly** -  Tests that should only be run manually. They should not be run as part of any CI/CD

**TS4UWP** - A superset of all tests related to the UWP output
**TS4WPF** - A superset of all tests related to the WPF output
**TS4WinUICPP** - A superset of all tests related to the WinUI (C++) output
**TS4WinUICS** - A superset of all tests related to the WinUI (C#) output


It is expected a new (updated) "Nightly" CI/CD process will be defined and run all tests apart from the "Release" and "ManualOnly" tests.



Traits with the key "Type" are also defined. These are a hangover from the VS2019 version. Their consistency is not guaranteed. It may be appropriate to remove these if they're not needed.

Previously, traits with the key "ExecutionSet" were defined but these have been removed to simplify the options used.


 CI/CD scripts will need to be updated to use the new "Group" traits.

mrlacey:
I think it's there.
Obviously with so many changes, reviewing the details here is not going got be a good use of time.
Instead, I think the best thing to do will be to get the PR and check everything builds and runs as expected.

The automated PR checks **will not work** because some of the files they look for no longer exist.

Please see the above comments to explain and summarize the key aspects of the changes.

Questions welcome.

mikebattista (approved):

mikebattista:
Thanks for this massive update and restructuring! Doesn't need to be perfect starting out, and we can continue iterating once we all start working with the new project structure.

PR #4415, opened 2022-03-25
remove some old commented out code

mrlacey:
this is just a bit of tidying up in the code, so I can make a PR and see the new checks being run against it. :)

### DIFF
--- a/code/SharedFunctionality.Core/Templates/ITemplateInfoExtensions.cs
+++ b/code/SharedFunctionality.Core/Templates/ITemplateInfoExtensions.cs
@@ -492,15 +492,6 @@ namespace Microsoft.Templates.Core
 
         private static string GetConfigDir(ITemplateInfo ti)
         {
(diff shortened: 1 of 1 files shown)


mrlacey:
Looks like this is failing because the old Template Tests haven't been removed. They're failing based on things that are no longer part of the project.
These have been superseded by the updated PR tests.

mikebattista (approved):

mikebattista:
The new PR pipeline passed so is working as expected.

Can we get rid of the [Template Tests (Full)](https://github.com/microsoft/TemplateStudio/blob/main/_build/pipelines/templates.tests.full.yml) pipeline now? What about the [OneByOne](https://github.com/microsoft/TemplateStudio/blob/main/_build/pipelines/templates.tests.onebyone.yml) pipeline?

Also should we consolidate the ci.yml and pr.yml pipelines into just the pr.xml pipeline now?

PR #4417, opened 2022-03-29
Addressed build errors and warnings in the CI pipeline

mikebattista:
# PR checklist

**Quick summary of changes**
See commits for details. Addressed build errors and warnings in the CI pipeline.

| UWP              | WPF              | WinUI            |
| :--------------- | :--------------- | :----------------|
| No | No | Yes |

Once the PR is finished, then yes.

N/A

**If you've changed the UI:**

  - Be sure you reviewed the [Template Verification Checklist](https://github.com/microsoft/WindowsTemplateStudio/wiki/Checklist:-Template-Verification).
  - Be sure it's included in the list on this [UWP](https://github.com/microsoft/WindowsTemplateStudio/blob/dev/docs/UWP/getting-started-endusers.md) or [WPF](https://github.com/microsoft/WindowsTemplateStudio/blob/dev/docs/WPF/getting-started-endusers.md) getting started doc.

N/A

**Have docs been updated?**
N/A

N/A


code/TemplateStudioForUWP/Templates/_catalog/cs-CZ.frontendframeworks.json
```diff
@@ -2,16 +2,16 @@
   {
     "name": "MVVMToolkit",
     "displayName": "MVVM Toolkit",
-    "summary": "Moderní, rychlá, modulární knihovna MVVM a součást sady nástrojů Windows Community Toolkit."
+    "summary": "ModernÃ­, rychlÃ¡, modulÃ¡rnÃ­ knihovna MVVM a souÄÃ¡st sady nÃ¡strojÅ¯ Windows Community Toolkit."
```
mrlacey, line 5: This looks broken.
What has happened to all these localized files?

mrlacey (changes requested):
Fix the encoding issue with localized files then all looks good

code/TemplateStudioForUWP/Templates/_catalog/cs-CZ.frontendframeworks.json
mikebattista: I'm not sure. I hit encoding issues in the tests, so I ran the script that's supposed to fix all encoding issues. Is https://github.com/microsoft/TemplateStudio/blob/main/_utils/Set-All-Text-File-Template-Encodings.ps1 broken?

mikebattista: Also it seems every time I touch a file I get encoding issues. Can we enforce the desired encoding via .editorconfig or some other mechanism so files are always saved with the right encoding?

mikebattista: I'm using VS, and VS Code sometimes for some things. I had never opened the json or md files, though, the script just updated them when I ran it. I'll just revert all changes to the json and md files for now.

mrlacey (approved):

mikebattista:
UWP tests are all passing with 1 "Other" that's causing the test pass to fail. Build output is useless and doesn't indicate what test has problems or provide a useful error message.

mrlacey:
> UWP tests are all passing with 1 "Other" that's causing the test pass to fail. Build output is useless and doesn't indicate what test has problems or provide a useful error message.

I think it's the Core tests that are the cause of this and have raised a follow-up issue.

mikebattista:
Regarding the encoding issues, I think what happened is I ran the script in PowerShell Core first, which throws a bunch of errors, but actually modifies all the json and md files that were problematic. Running the script in Windows PowerShell works as expected.

If the script is Windows PowerShell only, should we add a check to block execution if running under PowerShell Core? Should we update the script to work properly in PowerShell Core?

mikebattista:
I think the below would make it PowerShell Core-compatible but wouldn't work in Windows PowerShell:

```diff
-   [byte[]]$byte = Get-Content -Encoding byte -ReadCount 3 -TotalCount 3 -Path $Path
+   [byte[]]$byte = Get-Content -AsByteStream -Raw -Path $Path
```

PR #4428, opened 2022-04-05
Updated the release pipeline

mikebattista:
The pipeline produces production signed release builds of the extensions.

It does not run any tests right now.


_build/pipelines/release.yml
```diff
@@ -0,0 +1,191 @@
...
+        BuildConfiguration: ReleaseUWP
+        BuildPlatform: Any CPU
+
+  pool:
+    vmImage: windows-2022
+
+  steps:
+  - checkout: self
+    clean: true
+    submodules: recursive
```
mrlacey, line 43: shouldn't need any submodules - the one we did use (CoreTS) isn't used any more.

mrlacey (approved):
This is not something that I can use and test locally as it's dependent on some Microsoft signing infrastructure.
Approving on the basis you've run it and it works. :)

PR #4435, opened 2022-04-05
Nightly build pipeline

mikebattista:
Haven't tested yet but wanted to share for review. Deltas from CI pipeline:

* Nightly trigger instead of CI/PR triggers
* No timeouts on the jobs
* Removed the Release build jobs
* Changed the Group in the traits to Build*

Will test once your test refactoring merges to main.


mikebattista:
@mrlacey the UWP tests failed twice with:

```
,##[error]We stopped hearing from agent Azure Pipelines 10. Verify the agent machine is running and has a healthy network connection. Anything that terminates an agent process, starves it for CPU, or blocks its network access can cause this error. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610
```

The first run also went over 6 hours which is the max time limit for a job on a Microsoft-hosted agent. See https://docs.microsoft.com/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts.

We saw these issues with the UWP tests before the refactor as well. How do you want to handle? Need to improve the reliability and ensure each job we run is under 6 hours, so we may need to split up the UWP tests if they're expected to run for longer.

mrlacey:
I've added a change to increase the parallelization of the tests. This _should_ greatly reduce execution time.

How can I test this running in the real (or an equivalent) environment?

mikebattista:
> I've added a change to increase the parallelization of the tests. This _should_ greatly reduce execution time.
> 
> How can I test this running in the real (or an equivalent) environment?

I kicked off a build. If you have access to Azure DevOps, you can always set up your own pipelines against the yaml files in the repo to run them yourself. The only one you wouldn't be able to run is the release pipeline that relies on private signing infrastructure.

mrlacey:
I'm getting some weird errors when running them in parallel and will continue to investigate.

To unblock:
- Disable the running of the UWP tests in this script
- Create a new issue to get them running correctly and assign it to me
- I'll look at this separately.

_build/ParallelTestExecution.ps1
```diff
@@ -21,7 +21,7 @@ workflow pbatch {
 
     foreach -parallel ($trait in $traits) {
         $outData = InlineScript {
-            . $Using:testrunner $Using:testLibrary -xml "$Using:outputDir\TEST-$Using:trait-Result.xml" -parallel none -trait $Using:trait 
+            . $Using:testrunner $Using:testLibrary -xml "$Using:outputDir\TEST-$Using:trait-Result.xml" -parallel all -trait $Using:trait
```
mrlacey, line 24: revert this then good to go

mrlacey, line 24:
```suggestion
            . $Using:testrunner $Using:testLibrary -xml "$Using:outputDir\TEST-$Using:trait-Result.xml" -parallel none -trait $Using:trait 
```

mrlacey (approved):
with reversion comment

Issue #4438, opened 2022-04-06
QuickWin: improve PR build times


mrlacey:
### Is your feature request related to a problem? Please describe.

The build and test processes we run for each PR take a long time.
Ideally, these should be as quick as possible.

There are (at least) four things we could do to improve these:
- run fewer checks [not desirable]
- rewrite the tests (the biggest contributor to the overall time taken) to run faster [Not a quick fix] 
- Only run tests on the parts of the code that have changed [eventually we'll get to this, but it's a lower priority]
- look for quick fixes to what we currently have [this! do this now!]

### Describe the solution you'd like

I think I can make the `Test Shared Core` and `Test Shared UI` actions faster by adding a new Solution Filter just for them so that the task doesn't need to build things that won't be tested.

Arguably I could add two new solution filters (one for each) but I'd like to keep the number we have small and hopefully manageable. (This can be reviewed at a later date if needed.)

### Describe alternatives you've considered

see above

### Additional context

I think this can save 5-10 minutes and so is definitely worth trying.

### Applies to the following platforms:

- [ ] WinUI
- [ ] WPF
- [ ] UWP

mrlacey:
I expect I'll be motivated to look at the conditional execution of individual jobs next time I'm anxiously waiting for a set of PR checks to run 😉

mikebattista:
Merged with #4439.

PR #4439, opened 2022-04-06
Add solution filter for shared tests

mrlacey:
to be used in the PR pipeline
For #4438

# PR checklist

**Quick summary of changes**

- [ ] WinUI
- [ ] WPF
- [ ] UWP

**If you've changed the UI:**

**Have docs been updated?**

### DIFF
--- a/_build/pipelines/ci.yml
+++ b/_build/pipelines/ci.yml
@@ -102,10 +102,11 @@ jobs:
         Moniker: WPF
         BuildConfiguration: DebugWPF
         BuildPlatform: Any CPU
-      UWP:
-        Moniker: UWP
-        BuildConfiguration: DebugUWP
-        BuildPlatform: Any CPU
+      # Reenable as part of #4441
+      #UWP:
+      #  Moniker: UWP
+      #  BuildConfiguration: DebugUWP
+      #  BuildPlatform: Any CPU
 
   pool:
     vmImage: windows-2022
--- a/_build/pipelines/ci.yml
+++ b/_build/pipelines/ci.yml
@@ -153,14 +153,14 @@ jobs:
 
   strategy:
     matrix:
-      Core:
-        Moniker: Core
-        FrameworkDirectory: net472
-        BuildConfiguration: DebugALL
-        BuildPlatform: Any CPU
       UI:
         Moniker: UI
         FrameworkDirectory: .
+        BuildConfiguration: DebugALL
+        BuildPlatform: Any CPU
+      Core:
+        Moniker: Core
+        FrameworkDirectory: net472
         BuildConfiguration: DebugALL
         BuildPlatform: Any CPU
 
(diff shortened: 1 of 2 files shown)


mrlacey:
That's an even bigger improvement than I was expecting!

Although these jobs may be running in parallel and so the impact of this change may be reduced.

Need to allow for some variations in build times on the server (because of load & _"reasons"_) but even still this is better.
Less time waiting for checks to complete. 😃 


**Before** (random sample):

13m 13s & 10m 18s

**After** (build for this PR):

4m 15s & 3m 24s

mrlacey:
This [failed on one attempt](https://winappstudio.visualstudio.com/WTS/_build/results?buildId=38340&view=logs&j=37d10d5d-e62d-5f4f-b769-ae2539dbd684&t=532deab7-1da9-5ec5-b782-b692fc07a8c3) due to an intermittent error with the nuget restore task.
@mikebattista how do you feel about adding automatic retries for things like this?

mikebattista:
Conditional execution of build/test based on what changed we can look at after we ship the initial VS 2022 versions.

What we need for that, though, is a spec for what changes in the repo require what platforms and tests to be run.

mikebattista:
You know, another quick win here would be to not run the UWP tests for CI given that (1) we don't expect many changes there going forward and (2) it takes the majority of the time (1 hour+ compared to 5-20 minutes for everything else).

If we assume the changes to UWP will be infrequent and that the nightly test passes would catch anything egregious, we can disable that from CI.

mikebattista:
> @mikebattista Why change the order of the Shared matrix after you reverted the Test order change?

It's purely cosmetic.

_build/pipelines/ci.yml
```diff
@@ -102,8 +102,8 @@ jobs:
         Moniker: WPF
         BuildConfiguration: DebugWPF
         BuildPlatform: Any CPU
-      # Reenable as part of #4441
-      #UWP:
+      # #4441
```
mrlacey, line 105: Why remove additional context for this issue number?
If all I see is an issue number and I don't remember why it's there (I won't,) if I'm looking at this file and wonder about this commented out section I then have to go and look at the referenced issue to see why that's the case.
Does the referenced issue explain why it's commented out? Does it explain when or why to reenable it? Is it important? Is it just there for information purposes only? Does it directly relate to the commented-out lines below?

I'm all for standardizing formatting and having after the character that indicates a comment, but removing contextual information that is intended to aid in future code maintenance (for the sake of "cosmetic" improvements) doesn't feel like a useful change.

**Not just me but anyone who looks at this file.** - Others are even more likely to need to go and look this up.

mikebattista: "Reenable as part of #4441" also isn't enough context to answer the questions you raised. If you think looking at commented lines here is confusing, I'd rather just remove the whole UWP section, but I prefer the way it is now with comments and an issue number for reference.

mikebattista (approved):

Issue #4487, opened 2022-05-10
Use expression body for methods


mikebattista:
### Describe the bug

We have inconsistent application of expression bodies on methods, and default CA rules do not complain about this. With .editorconfig, we can enable warnings for where expression bodies can be used on methods. We should apply expression bodies on methods where possible for consistency and to apply and promote latest language features.

Expression bodies on properties was addressed via #4484. #4486 covers addressing issues raised by the default CA warnings. Since using expression bodies on methods is not enabled in the default CA warnings, filing a separate issue to track since I believe we should make it part of our guidelines.

### Applies to the following platforms:

- [X] WinUI
- [ ] WPF
- [ ] UWP

### About your setup

 - Visual Studio Version:  
 - Template Studio Wizard Version: 
 - Windows Build:

mikebattista:
Merged with #4490.

Issue #4486, opened 2022-05-09
Resolve Code Analysis messages in WinUI templates


mikebattista:
### Describe the bug

Run Code Analysis and enable Messages in the Error List to see recommendations from the default CA rules. Most of these involve using new C# language features.

We should resolve these so there are no CA warnings out-of-the-box.

### To Reproduce

Run Code Analysis and enable Messages in the Error List to see recommendations from the default CA rules.

### Applies to the following platforms:

- [X] WinUI
- [ ] WPF
- [ ] UWP

### About your setup

 - Visual Studio Version:  
 - Template Studio Wizard Version: 
 - Windows Build:

mikebattista:
Merged with #4490.

PR #4490, opened 2022-05-12
Fixed CA warnings in the WinUI templates.

mikebattista:
# PR checklist

**Quick summary of changes**
Added .editorconfig to generated projects with Warning severity on rules modified from the default rules. Generate a project, run CA, and you should only see warnings for #4487 (https://github.com/microsoft/TemplateStudio/pull/4490#issuecomment-1127041242). Could revert back to Suggestion but I found some rules like expression bodied syntax on accessors weren't triggered everywhere unless the severity was increased.

#4486
#4487

- [X] WinUI
- [ ] WPF
- [ ] UWP

**If you've changed the UI:**

**Have docs been updated?**


mikebattista:
Test failure is #4491. Not sure what's going on there.

mrlacey:
Where did the contents of the editorconfig come from? Are they all defaults?
Why the switch to the beta version of the StyleCop analyzers? - What does the new version have that's needed?

mikebattista:
> Where did the contents of the editorconfig come from? Are they all defaults? Why the switch to the beta version of the StyleCop analyzers? - What does the new version have that's needed?

The original .editorconfig was added through VS's Add New Item menus with all default values. I then changed the items that now have severity = warning as mentioned in the PR description. The rules I changed were based on the original errors that CA was reporting and that were fixed with the changes here, and that we should enforce going forward. I also enabled "expression bodies on methods" since my proposal is that we adhere to that, though that is tracked separately by #4487 since it's debatable and a more significant change.

There was a bug in the stable version of StyleCop that was conflicting with the rule to use `new()` when the type could be implied. StyleCop was complaining that new should have a space after it (https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3459). Suppressing the rule would have suppressed other legitimate checks.

mikebattista:
With https://github.com/microsoft/TemplateStudio/pull/4490/commits/0aca68350ff93ed0997171fac88f84a8f192c7da I've fixed what I believe are all the expression body on method CA warnings that we should fix. In short, I don't think we should require it everywhere it can be used, some judgment was involved:

* I reduced the .editorconfig rule to only warn when the method body is a single line, so that excludes some applications like in [Json.cs](https://github.com/microsoft/TemplateStudio/blob/main/code/TemplateStudioForWinUICs/Templates/Ft/SettingsStorage/Param_ProjectName.Core/Helpers/Json.cs). Bodies more complicated than a single line start raising formatting questions and are more likely to expand to more than one statement in the future. I'd rather use traditional bodies in these cases and .editorconfig has an option specifically to exclude these cases.
* Any Get/Set type methods that have single line bodies should use expression bodies. See [NavigationViewHeaderBehavior](https://github.com/microsoft/TemplateStudio/blob/0aca68350ff93ed0997171fac88f84a8f192c7da/code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Behaviors/NavigationViewHeaderBehavior.cs). This mirrors the rules for expression bodies on properties and accessors.
* Local consistency matters. If a file has 9 methods with classic method bodies and one that could use expression bodies, I'd rather use a classic method body for the 10th to maintain local consistency in the file.
* The CA rule to use expression bodies on methods is not enabled in the default CA rules. After going through the codebase and making calls like above, I understand why and think we should keep any warnings for this off by default in the .editorconfig.

In terms of formatting, the whole definition should be on one line. If the line needs to wrap, then the first line should end with `=>` and the next line should contain the expression body. I fixed some formatting issues as part of this change as well.

https://github.com/microsoft/TemplateStudio/pull/4490/commits/0aca68350ff93ed0997171fac88f84a8f192c7da lists the issues I fixed. To see the remaining issues I didn't fix, create a NavigationView project with all pages, then run CA. You should see the remaining expression body on method warnings that I decided not to fix. If you agree with the won't fixes, then we should disable the rule in the .editorconfig to revert back to the default behavior.

mrlacey:
#4497 should unblock this

code/TemplateStudioForWinUICs/Templates/Ft/SettingsStorage/Param_ProjectName/Services/LocalSettingsServicePackaged.cs
```diff
@@ -9,7 +9,7 @@ public class LocalSettingsServicePackaged : ILocalSettingsService
     {
         public async Task<T> ReadSettingAsync<T>(string key)
         {
-            object obj = null;
+            object obj;
 
             if (ApplicationData.Current.LocalSettings.Values.TryGetValue(key, out obj))
```
mrlacey, line 14: WHy not fully inline the declaration?

code/TemplateStudioForWinUICs/Templates/Ft/ThemeSelection/Services/ThemeSelectorService.cs
```diff
@@ -46,20 +46,19 @@ public async Task SetRequestedThemeAsync()
 
         private async Task<ElementTheme> LoadThemeFromSettingsAsync()
         {
-            ElementTheme cacheTheme = ElementTheme.Default;
-            string themeName = await _localSettingsService.ReadSettingAsync<string>(SettingsKey);
+            var themeName = await _localSettingsService.ReadSettingAsync<string>(SettingsKey);
 
-            if (!string.IsNullOrEmpty(themeName))
+            if (Enum.TryParse(themeName, out ElementTheme cacheTheme))
```
mrlacey, line 51: Does this definitely still work if `themeName` is null (or empty)?

mikebattista: Yes.

code/TemplateStudioForWinUICs/Templates/Ft/SettingsStorage/Param_ProjectName/Services/LocalSettingsServicePackaged.cs
mikebattista: Fixed. Thanks.

mrlacey (approved):
It would be good if, one day, we can also do away with the need to have the StyleCop checks entirely and do it all with .editorconfig.

Issue #4463, opened 2022-04-22
Update UX to match WinUI design guidelines


mikebattista:
### Is your feature request related to a problem? Please describe.

Need to update UX for latest WinUI design guidelines (margins, layouts, styles, etc.).

* [x] Add unique icons per page template in NavigationView (#4488)
  * [x] Page for Blank
  * [x] Globe for WebView
  * [x] List for ListView
  * [x] GridView for ContentGrid
  * [x] PageMarginLandscapeWide for DataGrid
* [x] Remove unnecessary Page, Frame, and Grid Styles and BackgroundColors
* [x] Add custom titlebar required for dark theme and Mica support
* [ ] Add Mica support (moved out to 5.2)
* [x] MenuBar Settings page should follow Notepad's Settings page instead of a flyout

### Applies to the following platforms:

- [X] WinUI
- [ ] WPF
- [ ] UWP

mrlacey:
At the very least, this issue needs a link to the guidelines. 
A breakdown of what needs changing is also required so that progress can be monitored and it's possible to evaluate when it's done and that the changes made match what is required.

mikebattista:
There's an internal design review in progress and will list the action items here once that is resolved.

mrlacey:
For additional background information. The NavView icons were previously, deliberately, all the same, to highlight that they should be changed.
Was any consideration given to the possibility of this now being seen as advice or a recommendation that these pages _should_ use these specific icons?

mikebattista:
The TODO comments are still there and have been updated to guide the developer towards how to use the latest Fluent icons, but the out-of-box experience has always looked a bit broken to me, and the designers agreed. Differentiating the icons looks a lot more polished.

mikebattista:
All changes have been merged.

PR #4505, opened 2022-05-18
Custom titlebar

smorin4:
Added custom titlebar required for dark theme 

Issue: #4463 

- [x] WinUI
- [ ] WPF
- [ ] UWP


mikebattista (commented):
Need to localize "App Title" by setting AppTitleBarText.Text to "AppDisplayName".GetLocalized() in the code behind.

mikebattista (commented):
Styles are off but we can revisit as part of reviewing the remove custom styles PR.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Views/ShellPage.xaml.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Input;
+using Param_RootNamespace.Helpers;
```
mikebattista, line 3: Please sort alphabetically. Helpers should come between Contracts and ViewModels.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.MenuBar/Views/ShellPage.xaml
```diff
@@ -21,10 +21,14 @@
...
         <Grid>
             <Grid.RowDefinitions>
-                <RowDefinition Height="Auto" />
+                <RowDefinition Height="32" />
+                <RowDefinition Height="48" />
                 <RowDefinition Height="*" />
             </Grid.RowDefinitions>
-            <MenuBar VerticalAlignment="Top">
+            <Grid Grid.Row="0" x:Name="AppTitleBar">
+                <TextBlock x:Name="AppTitleBarText" Text="App Title" VerticalAlignment="Center" TextWrapping="NoWrap" Style="{StaticResource CaptionTextBlockStyle}" Margin="28, 0, 0, 0"/>
```
mikebattista, line 29: We should add an icon here as shown in https://docs.microsoft.com/en-us/windows/apps/develop/title-bar?tabs=winui3#title-bar-content-and-drag-regions. There may be one in the repo we can use.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Views/ShellPage.xaml.cs
mrlacey: There's a postaction that sorts all namespaces in the generated code. 
If we want to ensure they're presorted in the templates too, we should add a test/check for this as part of the minimal tests.

mrlacey:
> Added custom titlebar required for dark theme
> 
> Issue: #4463

What are the actual requirements here? The linked issue doesn't have any details.
Without these details (and knowing what should be being done) it's impossible to review the code and confirm it is doing the right thing (and only the things it's supposed to.)

mrlacey:
> Need to localize "App Title" by setting AppTitleBarText.Text to "AppDisplayName".GetLocalized() in the code behind.

Can it not be done through the XAML? Seems odd that this needs to be done in codebehind.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Views/ShellPage.xaml.cs
mikebattista: The test is not a priority right now for 5.1 but we should make the change here.

mikebattista:
> > Added custom titlebar required for dark theme
> > Issue: #4463
> 
> What are the actual requirements here? The linked issue doesn't have any details. Without these details (and knowing what should be being done) it's impossible to review the code and confirm it is doing the right thing (and only the things it's supposed to.)

We're implementing https://docs.microsoft.com/en-us/windows/apps/develop/title-bar?tabs=winui3#full-customization.

mikebattista:
> > Need to localize "App Title" by setting AppTitleBarText.Text to "AppDisplayName".GetLocalized() in the code behind.
> 
> Can it not be done through the XAML? Seems odd that this needs to be done in codebehind.

Doesn't need to be, but we already have AppDisplayName as a resource in the resw and this is the equivalent of how we're setting the Window title today (in App.xaml.cs).

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.MenuBar/Views/ShellPage.xaml
mikebattista: We use an Assets folder already when MSIX is enabled. We should put this in the Assets folder rather than Images.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.MenuBar/Views/ShellPage.xaml.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Input;
+using Param_RootNamespace.Helpers;
```
mikebattista, line 3: Update sort order here as well.

mrlacey:
> > > Need to localize "App Title" by setting AppTitleBarText.Text to "AppDisplayName".GetLocalized() in the code behind.
> > 
> > 
> > Can it not be done through the XAML? Seems odd that this needs to be done in codebehind.
> 
> Doesn't need to be, but we already have AppDisplayName as a resource in the resw and this is the equivalent of how we're setting the Window title today (in App.xaml.cs).

I believe the title is only set in app.xaml.cs because a framework bug that prevents setting the title in XAML.

mikebattista:
> > > > Need to localize "App Title" by setting AppTitleBarText.Text to "AppDisplayName".GetLocalized() in the code behind.
> > > 
> > > 
> > > Can it not be done through the XAML? Seems odd that this needs to be done in codebehind.
> > 
> > 
> > Doesn't need to be, but we already have AppDisplayName as a resource in the resw and this is the equivalent of how we're setting the Window title today (in App.xaml.cs).
> 
> I believe the title is only set in app.xaml.cs because a framework bug that prevents setting the title in XAML.

Ok. Either way, we're already using code behind to set other properties of the titlebar and I'd rather not duplicate the AppDisplayName in resw just so we can use x:Uid here. I'd prefer the code behind approach we have.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Views/ShellPage.xaml.cs
```diff
@@ -48,5 +52,6 @@ private static void OnKeyboardAcceleratorInvoked(KeyboardAccelerator sender, Key
             var result = navigationService.GoBack();
             args.Handled = result;
         }
+
```
mrlacey, line 55: Why a blank line here?

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.MenuBar/Views/ShellPage.xaml
```diff
@@ -26,8 +26,8 @@
                 <RowDefinition Height="*" />
             </Grid.RowDefinitions>
             <Grid Grid.Row="0" x:Name="AppTitleBar">
-                <Image Source="design\Assets\Logo\TemplateStudio_16.png" HorizontalAlignment="Left" Width="16" Height="16" Margin="8,0"/>
-                <TextBlock x:Name="AppTitleBarText" Text="App Title" VerticalAlignment="Center" TextWrapping="NoWrap" Style="{StaticResource CaptionTextBlockStyle}" Margin="28, 0, 0, 0"/>
+                <Image x:Name="TitleBarImage" Source="/Assets/WindowIcon.png" HorizontalAlignment="Left" Width="16" Height="16" Margin="8,0"/>
```
mikebattista, line 29: We don't use this name anywhere in code behind so let's remove it for now. Otherwise I'd call it AppTitleBarImage to align with AppTitleBar above.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.MenuBar/Views/ShellPage.xaml
```diff
@@ -26,8 +26,8 @@
...
+                <TextBlock x:Name="AppTitleBarText" Text="App Title" VerticalAlignment="Center" TextWrapping="NoWrap" Style="{StaticResource CaptionTextBlockStyle}" Margin="32, 0, 0, 0"/>
```
mikebattista, line 30: Remove the Text property. We set this in code behind.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Views/ShellPage.xaml.cs
```diff
@@ -52,5 +52,6 @@ private static void OnKeyboardAcceleratorInvoked(KeyboardAccelerator sender, Key
...
```
mikebattista, line 55: Was this new line addition intentional? If not, please remove.

mrlacey:
> > > Added custom titlebar required for dark theme
> > > Issue: #4463
> > 
> > 
> > What are the actual requirements here? The linked issue doesn't have any details. Without these details (and knowing what should be being done) it's impossible to review the code and confirm it is doing the right thing (and only the things it's supposed to.)
> 
> We're implementing https://docs.microsoft.com/en-us/windows/apps/develop/title-bar?tabs=winui3#full-customization.

That link is to an optional customization. The description for this PR implies that this is "required for dark theme".
Which is it?

If it must be there for dark theme support, where are the details for that? Is that a bug? Is it documented somewhere? It sounds like something every developer should know. And why not only add this as part of the "Theme Selection" option, as that's where dark theme support gets added in the generated solution.

If this is purely to show how to customize the title bar, this isn't something that should be included by default or it risks becoming something that must be removed from the generated solution. A key part of the purpose behind the compositional nature of the wizard is that it's possible to add only the things in a generated solution that are actually wanted. This should be a separate selectable item.

mikebattista:
> > > > Added custom titlebar required for dark theme
> > > > Issue: #4463
> > > 
> > > 
> > > What are the actual requirements here? The linked issue doesn't have any details. Without these details (and knowing what should be being done) it's impossible to review the code and confirm it is doing the right thing (and only the things it's supposed to.)
> > 
> > 
> > We're implementing https://docs.microsoft.com/en-us/windows/apps/develop/title-bar?tabs=winui3#full-customization.
> 
> That link is to an optional customization. The description for this PR implies that this is "required for dark theme". Which is it?
> 
> If it must be there for dark theme support, where are the details for that? Is that a bug? Is it documented somewhere? It sounds like something every developer should know. And why not only add this as part of the "Theme Selection" option, as that's where dark theme support gets added in the generated solution.
> 
> If this is purely to show how to customize the title bar, this isn't something that should be included by default or it risks becoming something that must be removed from the generated solution. A key part of the purpose behind the compositional nature of the wizard is that it's possible to add only the things in a generated solution that are actually wanted. This should be a separate selectable item.


The default titlebar has issues. Custom titlebar is the leading practice which is required for dark theme as well as full window Mica support which we plan to add.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Views/ShellPage.xaml.cs
mikebattista: The CI tests failed due to StyleCop complaining about this new line. Please remove.

code/TemplateStudioForWinUICs/Templates/_comp/MT/Project.SplitView/Views/ShellPage.xaml.cs
mikebattista: Fixed.

mrlacey:
> > > > > Added custom titlebar required for dark theme
> > > > > Issue: #4463
> > > > 
> > > > 
> > > > What are the actual requirements here? The linked issue doesn't have any details. Without these details (and knowing what should be being done) it's impossible to review the code and confirm it is doing the right thing (and only the things it's supposed to.)
> > > 
> > > 
> > > We're implementing https://docs.microsoft.com/en-us/windows/apps/develop/title-bar?tabs=winui3#full-customization.
> > 
> > 
> > That link is to an optional customization. The description for this PR implies that this is "required for dark theme". Which is it?
> > If it must be there for dark theme support, where are the details for that? Is that a bug? Is it documented somewhere? It sounds like something every developer should know. And why not only add this as part of the "Theme Selection" option, as that's where dark theme support gets added in the generated solution.
> > If this is purely to show how to customize the title bar, this isn't something that should be included by default or it risks becoming something that must be removed from the generated solution. A key part of the purpose behind the compositional nature of the wizard is that it's possible to add only the things in a generated solution that are actually wanted. This should be a separate selectable item.
> 
> The default titlebar has issues. Custom titlebar is the leading practice which is required for dark theme as well as full window Mica support which we plan to add.

Does that include adding a custom image too?

mikebattista:
The custom titlebar docs I linked recommend adding a WindowIcon yes. We are following the documented patterns. Developers can replace with their own icons.

mrlacey:
> The custom titlebar docs I linked recommend adding a WindowIcon yes. We are following the documented patterns. Developers can replace with their own icons.

That linked document says that customizing the title bar is optional.

Is the dark theme issue you refer to this one? https://github.com/microsoft/microsoft-ui-xaml/issues/5408

mikebattista:
> > The custom titlebar docs I linked recommend adding a WindowIcon yes. We are following the documented patterns. Developers can replace with their own icons.
> 
> That linked document says that customizing the title bar is optional.
> 
> Is the dark theme issue you refer to this one? [microsoft/microsoft-ui-xaml#5408](https://github.com/microsoft/microsoft-ui-xaml/issues/5408)

Yes exactly. Custom titlebar is required for dark theme because of that bug. It is also required for Mica by design. So based on the design guidelines we want to comply with, it is required.

mrlacey:
> Yes exactly. Custom titlebar is required for dark theme because of that bug. It is also required for Mica by design. So based on the design guidelines we want to comply with, it is required.

So what about when that bug is fixed?
I also presume that the need to use a custom title bar to use a Mica background is a temporary issue. 

If the generated code includes something to work around a known issue it would be good to include appropriate comments in the generated code to explain this. (This is also what has historically been done in Template Studio.)


For clarity, I assume the design doc referred to above is https://docs.microsoft.com/en-us/windows/apps/design/style/mica#when-to-use-mica

mikebattista:
> > Yes exactly. Custom titlebar is required for dark theme because of that bug. It is also required for Mica by design. So based on the design guidelines we want to comply with, it is required.
> 
> So what about when that bug is fixed? I also presume that the need to use a custom title bar to use a Mica background is a temporary issue.
> 
> If the generated code includes something to work around a known issue it would be good to include appropriate comments in the generated code to explain this. (This is also what has historically been done in Template Studio.)
> 
> For clarity, I assume the design doc referred to above is https://docs.microsoft.com/en-us/windows/apps/design/style/mica#when-to-use-mica

This is required for more than just dark theme. Mica will require this. The link you shared is for WinUI 2. WinUI 3 has same design principles but is harder to implement right now in 1.1. We're also updating NavigationView designs based on designer feedback to merge the control with the custom title bar.

mikebattista:
Need to fix #4517 then ready to merge.

mikebattista (approved):

Issue #4528, opened 2022-06-03
Add MSTest support


mikebattista:
### Is your feature request related to a problem? Please describe.

https://github.com/microsoft/testfx/pull/1071

### Applies to the following platforms:

- [X] WinUI
- [ ] WPF
- [ ] UWP

PR #4549, opened 2022-06-21
Add MSTest support

smorin4:
# PR checklist

Added MSTests option to wizard. Generates MSTest project that references main/core projects and includes scaffolding + stubs for testing the view models used by different added pages. Currently supports MSIX packaged and unpackaged projects.

Issue: #4528 

- [x] WinUI
- [ ] WPF
- [ ] UWP

Currently contains a bug with ResourceLoader, which is responsible for obtaining app settings data for the Settings View Model. The test project is unable to retrieve the resource file from the main project. Therefore, when all pages are added and their corresponding tests are run, all will pass except for the SettingViewModelTest.

Addition to wizard:
Testing Screenshot


mikebattista:
With the edit project menu now restored, let's make sure the MSTest project is offered under the test menu.

mikebattista:
@smorin4 the tests are discovered for me but don't run in the generated project. Could you take a look?

Also could you test with all the different architectures that the app supports (e.g. x86, x64, etc.)? I think there is some configuration missing in the test project to support anything other than x86.

smorin4:
@mikebattista The tests are not running because the package references for MSTest.TestAdapter & MSTest.TestFramework are generated using version 2.2.8 instead of 2.2.10. Changing these to the latter fixes the problem.

As for the architecture support, you are correct that the test project only supports x86. Looking into solutions for this.

mikebattista:
Remaining work:

- [ ] Test framework-dependent scenarios

mikebattista:
Filed https://github.com/microsoft/TemplateStudio/issues/4576 to track main remaining issue.

mikebattista (approved):

Issue #4442, opened 2022-04-06
Deploy is unchecked in the build configurations for packaged projects


mikebattista:
### Describe the bug

This requires the developer to manually check Deploy in the build configuration or experience a failure the first time they try to F5.

This issue was fixed with the inbox Windows App SDK templates, but after some testing, Template Studio may not be benefitting from the fixes that were made for that, possibly related to its use of the .NET templating engine and how it composes templates together.

### To Reproduce

 1. Create a project with MSIX
 2. F5
 3. Observe the error

### Applies to the following platforms:

- [X] WinUI
- [ ] WPF
- [ ] UWP

### About your setup

 - Visual Studio Version:  
 - Template Studio Wizard Version: 
 - Windows Build:

mikebattista:
From within VS.

mikebattista:
MAUI fixed this in their project system last year via https://github.com/dotnet/maui/issues/2746.

Some other internal guidance on the fix, but this is based on the VS project system.

> Currently for both “Create New Project” and “Add New Project”,  the default state of “Deploy” flag for the new project is “calculated” by this logic:
>     If `<project DOES not support Deploying>`  then  it is set to FALSE.
>     Else if `<project DOES not implement hierarchy property “VSHPROPID_DefaultEnableDeployProjectCfg”>` then it is set to TRUE.
>   Else it is set to the value of “VSHPROPID_DefaultEnableDeployProjectCfg” property.
> 
> So, if you want the [new] project to end-up with “Deploy” checked, make sure that: 
> 1)	Either do not implement VSHPROPID_DefaultEnableDeployProjectCfg (aka return E_NOTIMPL) or if implemented, it returns “VARIANT_TRUE”. 
> 2)	Ensure that project “supports” deployment which means:
> a)	Project does implement IVsDeployableProjectCfg.
> b)	And IVsDeployableProjectCfg->QueryStartDeploy(0, &fSupported, &fReady); succeeds and sets fSupported to true.
>

mikebattista:
@mrlacey do you know how the above guidance might apply here?

mrlacey:
> @mrlacey do you know how the above guidance might apply here?

I'll dig deeper later, but it looks like the MAUI notes on the fix don't apply as they're for the project system. I'll start by looking at what changed in the "official" templates:
https://github.com/microsoft/WindowsAppSDK/commit/b419d6d9d551f8e083f97f647f0e51f654a319a3
[...]

PR #4562, opened 2022-07-06
Work-around for deploy not enabled in new WinUI projects

mrlacey:
# PR checklist

**Quick summary of changes**

Work-around for deployment issue of newly generated WinUI3 projects.
But a full fix for this when using VSEmulator.

#4442

- [x] WinUI
- [x] WPF
- [x] UWP

NOTE. This is a change made for WinUI projects but affects everything as it's in shared logic.

This is just a (hopefully) temporary work-around for the problem. 
More investigation is needed to identify the true cause of the issue.

The delay of three seconds is my best guess of how long to wait. 
Because I don't know exactly the cause of the issue, I don't know how long it is necessary to wait, it may also be connected to the machine that VS is running on and so want to allow sufficient time for machines under greater resource constraint than I've tested on.

The question that really needs to be asked is: Is a six-second delay in project generation better than relying on a person manually changing the generated project configuration? (And knowing that they have to!?)

Yes.

n/a

**If you've changed the UI:**

n/a

n/a

Suggest leaving the original issue open to continue looking into this.

**Have docs been updated?**

n/a

n/a


mrlacey:
@mikebattista assigning to you for review as this is related to something I think you'll want in the next release--which I assume you want to be soon based on the recent 1.1.2 release of WinAppSDK.

mikebattista:
I will take a look. Thanks. Would be nice to include the next release.

mikebattista:
@mrlacey was 3s arbitrarily chosen for the delay or did you test and discover that 3 seconds was the minimum to be reliable enough?

mrlacey:
> @mrlacey was 3s arbitrarily chosen for the delay or did you test and discover that 3 seconds was the minimum to be reliable enough?

As per above, 3 seconds is a best guess. 1 second wasn't always enough and I don't know how much device capabilities (processor speed, other load, etc.) play a factor. I only tested on two fairly high spec machines but know they are above what a lot of developers use.

> Also in the description you said 6 second delay, is that because this is applied to both the App and Core projects?

Yes, six seconds because it's a delay per project and the default is to have 2 projects in the generated solution. For generations adding lots of other projects (e.g. a web backend or test projects) but these are not widely included. 
We can revisit if this is a problem but I doubt it will be

mikebattista:
Given that this doesn't repro 100% and that the error dialog presented provides the fix when it does repro, I'm inclined not to take this workaround. The 6-9 second delay with 2-3 projects is a bit much and all extensions and projects would need to pay that penalty even though this problem only exists for the WinUI extension.

PR #4624, opened 2023-02-09
Update VS extension version to 5.3 and some minor fixes

pratikone:
# PR checklist

Updates VS extension versions to 5.3

** Minor fix **
Refer https://github.com/MicrosoftDocs/winrt-related/issues/19
In packaged apps, the app manifest was missing the changes made here
https://github.com/microsoft/WindowsAppSDK/pull/3206/files#

- [X] WinUI
- [X] WPF
- [X] UWP


mikebattista (approved):

code/TemplateStudioForWinUICpp/Templates/Features/MSIX/Param_ProjectName (Package)/Package.appxmanifest
```diff
@@ -12,7 +13,9 @@
     Publisher="CN=Param_UserName"
     Version="1.0.0.0" />
 
-  <Properties>
+   <mp:PhoneIdentity PhoneProductId="07ad151b-a614-4a13-aef0-5378d3f98e67" PhonePublisherId="00000000-0000-0000-0000-000000000000"/>
```
wjk, line 16: I don’t think we should put the same GUID in _all_ projects that are generated by this template. The Windows Store may give users grief if they update a package with a PhoneProductId that has already been used by another app.

code/TemplateStudioForWinUICs/Templates/Ft/MSIX/Package.appxmanifest
```diff
@@ -13,7 +14,9 @@
     Name="07ad151b-a614-4a13-aef0-5378d3f98e67"
     Publisher="CN=Param_UserName"
     Version="1.0.0.0" />
-
+    
+  <mp:PhoneIdentity PhoneProductId="07ad151b-a614-4a13-aef0-5378d3f98e67" PhonePublisherId="00000000-0000-0000-0000-000000000000"/>
```
wjk, line 18: Ditto.

wjk (changes requested):

code/TemplateStudioForWinUICpp/Templates/Features/MSIX/Param_ProjectName (Package)/Package.appxmanifest
mikebattista: These GUIDs get replaced during project generation.

Scottj1s: To remove any doubt/confusion, why not zero the guids out then?

mikebattista: template.json supports GUID replacements like https://github.com/microsoft/TemplateStudio/blob/6c3907aa00e08752a15b2394a01cc9d14222ede5/code/TemplateStudioForWinUICs/Templates/Ft/MSIX/.template.config/template.json#L30.

Each GUID listed gets replaced with a unique GUID. Given the templates were already following this pattern and this change just applied the replacement GUID to another element, I don't see a need to change, but you could 0 out the GUIDs and remove the autogeneration if you wanted.

PR #4627, opened 2023-02-10
Update ESRP CodeSign version to 2.0 for release pipeline

pratikone:
The release pipeline is failing because ESRP CodeSign version is too old and uses very old .Net core 2.0 dependency.
This PR updates it to 2.0 version so that it picks up modern .Net version and release pipeline becomes functional.


pratikone:
merging as i need to test if it is working properly. CI pipeline and product code are not affected.

PR #4682, opened 2023-05-31
Add Single Instance Template

Mysterious-Dev:
# PR checklist

**Quick summary of changes**

This PR add single instance template for `Template Studio for WinUI (C#)`.

I based my work on : https://blogs.windows.com/windowsdeveloper/2022/01/28/making-the-app-single-instanced-part-3/

**Which issue does this PR relate to?**

Close #4662 

**Applies to the following platforms:**

- [x] WinUI
- [ ] WPF
- [ ] UWP

PR #4693, opened 2023-06-20
Trim self-contained template

Mysterious-Dev:
# PR checklist

**Quick summary of changes**

This PR add Trim self-contained template for `Template Studio for WinUI (C#)`.

https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained

Like #4682, I don't included icon because I'm not a really good "designer" for that ^^.

- [x] WinUI
- [ ] WPF
- [ ] UWP


pratikone:
thanks. Will check this out

pratikone (commented):
Hi @Mysterious-Dev , once again, thank you for doing this. This looks like a high code quality change.
For the change, however, I discussed internally and decided to not take up this. Any new template adds to project complexity and thus effort to maintain it. So we need to justify adding a new project template.
These are the 2 reasons for not accepting this change: 
1. Template studio is primarily focused on beginners to get started with WinUI, without having to figure out config changes. This is an advanced concept by .Net and doesn't really add value to newbies using TS.
2. The changes to add trim is quite small (a change in .csproj) so any developer who needs it, can add it easily and unlike self contained, it is not a different app deployment type.

For any such big changes, I do suggest proposing it first (using the Issues tab) so that we can do the initial checks on whether we will be able to merge it with the product.
I do still appreciate the PRs you do for Template Studio and looking forward to more of them. 
Thank you and happy coding.

Mysterious-Dev:
@pratikone Hi ! I understand ^^. Well I think I can close this PR ?

pratikone:
sure. go ahead.

mrlacey:
While I appreciate the decision to gain wider feedback on a suggested change before accepting a PR, I'd like to push back on everything else discussed above.

- At what point did Template Studio switch to be "primarily focused on beginners"? Historically it strived to be useful to beginners and those with more experience.
- Including features as a way of raising awareness of them was always something that was a valid reason for adding items to the wizard. If someone was a beginner, then they may not be aware that there is a way of reducing the size of the generated assembly package. As beginners often want to compare a framework with other options, and generated assembly size is commonly used for this, it's probably good for beginners to be aware of.
- That this is a simple change is not an argument that has previously been used in assessing whether to include an item in the wizard. While it is a relatively simple change, it involves editing the file directly rather than using any specific UI. It's also possible to get in a situation where typing something incorrectly (due to limited intellisense) or pasting in the wrong place can lead to error messages that aren't very helpful. In that this change makes it so that all a person needs to do is click a button this strikes me as being a valuable addition.
- As this is probably the simplest addition that can be made to the project, I don't put a lot of value on the fact that this adds to the ongoing complexity and maintenance effort. It would be useful for the project to document how you justify the value of any new contribution and the effort required to maintain it so that any future contributor can assess whether a change/contribution is worth their time and effort.

